PR #9, opened 2015-10-01
Added sample link for AzureReceivers

garora:
Added sample link for AzureReceivers


garora:
@HenrikFrystykNielsen - Initiated the idea to implement ```webhook``` for ```Zendesk``` as per their [documentation](https://developer.zendesk.com/embeddables/docs/android/push_notifications_webhook#webhook-api). 

Tests still need to implement and sample required to test on actual environment.

I would appreciate if we can implement [core-api](https://developer.zendesk.com/rest_api/docs/core/introduction) and [triggers](https://developer.zendesk.com/rest_api/docs/core/triggers) too.

Request you to take a look and take this into your ```dev``` branch for further  implementation.

PR #19, opened 2015-10-21
Add a SqlStorage Provider

rposener:
With Only Memory Backed, or AzureTable storage options, we lose the ability to scale an application on IaaS. Particularly with the autoscale capabilities of Azure and the legacy components which prevent going full PaaS for our application.  Several other vendors are requesting that we develop integrations using webhooks, so we are really looking forward to this project to make that implementation simple.

If you think this has merit, I'm glad to continue to contribute to this effort and will help flesh out the tests, etc.  Perhaps it should also use SP's with simple SqlCommands instead?  Welcome to thoughts.


HenrikFrystykNielsen:
Thank you so much! Will have a look ASAP!

PR #30, opened 2015-11-29
Update param name to match IWebHookHandler

alindgren:
This class doesn't seem to use the param, so shouldn't effect anything but it should be consistently named.


HenrikFrystykNielsen:
Incorporated: e1b04c9463ae632f65b164a1ab390f62099f8411 Thanks!

PR #58, opened 2016-05-02
Added more samples

garora:
Added more samples


HenrikFrystykNielsen:
thanks @garora -- much appreciated!

PR #80, opened 2016-08-03
Corrected WebHook URI

garora:
Corrected WebHook URI, we do not required {id} part for Bitbucket WebHook URI.


HenrikFrystykNielsen:
Thanks! Merged as 14841dfdd3d93c6d040b25597fd5d3827524c3dc

PR #82, opened 2016-08-19
slack trigger can be optional 

ezlateva:
Hi! I've just come across a scenario when using the Slack webhook where Slack isn't sending either a 'trigger_word' or a 'command' parameter. This happens when you configure the outgoing webhook for a specific channel. In that case, trigger words are optional.

To make my receiver work in that scenario, I had to remove the requirement in code.

Thank you for your consideration of this pull request!
Elena


HenrikFrystykNielsen:
Awesome - thanks! Merged as bb39331cb9530ee99c9ce7ccaa2e547498799a5f

PR #95, opened 2016-11-13
Updated documentation link

garora:
Documentation links is changed and it redirects to asp.net core documentations, updated the new link to documentation.


HenrikFrystykNielsen:
Thanks but the fwlink URI now works again so I think we are good :)

PR #98, opened 2016-11-22
updates Azure package version

Kwal:


HenrikFrystykNielsen:
Thanks @Kwal but I think it's easier if people update the package themselves.

PR #116, opened 2017-02-13
Await for the response messages and then put responses into list.

petetheman79:
When a WebHook is added with an invalid Uri (host does not exist etc) by using the noecho feature so that the WebHook uri is not validated, when the WebHookSender sends the message an exception is thrown in AzureWebHookDequeueManager line 281 when waiting for all Tasks to complete. This causes the rest of the code to not be executed. This causes a situation where all messages are resent indefinitely causing duplicate messages to be transmitted.

The solution that I have is to remove the Task.WhenAll(requestTasks) and rather to await on the _parent._httpClient.SendAsync(request) to ensure that exceptions while posting are handled in the exception handling block.


HenrikFrystykNielsen:
Committed to dev branch as cb46ccaac71f9807f87d21af1ef4bc0783e6cb6c

syl20:
Hi, we experience some troubles with this bug and really need this fix :)
Can you publish the updated nuget package ?

dougbu:
Pre-release packages containing this fix are now available in the https://dotnet.myget.org/gallery/aspnetwebhooks feed.

syl20:
Thanks, we will test it with our development environment.

syl20:
Hi, we successfully test this package and it fixes our bug.
Now waiting for the release. Thanks.

PR #126, opened 2017-03-22
Reformatting readme

garora:
Somehow markdown formatting broke for headers level, reformatted and now these are appearing as these should.


HenrikFrystykNielsen:
Thanks -- much appreciated!

PR #132, opened 2017-04-26
400 Error during catch webhook(The ID 'evt_XXX' could not be resolved for an actual event)

Tryolki:
Added set the security protocols. The problem is caused by the fact that GetEventDataAsync receives an answer 403 during authorization step


HenrikFrystykNielsen:
Thanks! Accepted as a54d9fb030936b313365016327bb29303f16003c

PR #152, opened 2017-08-24
Add ASP.NET Core WebHooks support; one approach to #5

dougbu:
This approach maintains the controller / receiver / handler separation but switches everything from Web API to ASP.NET Core. For example, logging is done using `Microsoft.Extensions.Logging.ILogger` rather than a WebHooks-specific interface. This PR only adds `Microsoft.AspNetCore.Receivers.GitHub` and the two projects that depends on.

I'm working on a second approach that is significantly more ASP.NET Core-native. This will make more use of such MVC concepts as `IResourceFilter` and expose the `Controller` class to end users.

#### `IWebHookReceiver`
From: `Task<HttpResponseMessage> ReceiveAsync(string id, HttpRequestContext context, HttpRequestMessage request);`
To: `Task<IActionResult> ReceiveAsync(string id, HttpContext context, ModelStateDictionary modelState);`

#### `WebHookHandlerContext`
From: `public HttpResponseMessage Response { get; set; }`
To: `public IActionResult Result { get; set; }`

#### `WebHookHandler` sub-classes:
From: parameterless `base` constructor
To: pass the `ILoggerFactory`

From:` context.Response = context.Request.CreateErrorResponse(…)`
To: `context.Result = CreateErrorResult(…)`

#### `WebHookReceiver` sub-classes
My changes here amounted to a `WebHookReceiver` rewrite and will impact everyone extending the framework.

From: parameterless `base` constructor
To: pass in 7 different services

From: `void` helper methods that throw an `HttpResponseException`
To: either pass in a `ModelStateDictionary` (especially when an `IInputFormatter` may be used) or return an `IActionResult` (`null` if processing should continue)


src/Microsoft.AspNetCore.WebHooks.Common/Extensions/WebHooksHttpRequestExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+            if (type.Equals("application", StringComparison.OrdinalIgnoreCase) ||
+                type.Equals("text", StringComparison.OrdinalIgnoreCase))
+            {
+                return contentType.SubType.EndsWith("+xml", StringComparison.OrdinalIgnoreCase);
+            }
+
+            return false;
+        }
+
+        // ??? Should ASP.NET Core expose a similar HttpRequest, HttpContext or ConnectionInfo (extension) method?
```
dougbu: Comments like these i.e. including "???" are open questions that could or should be answered before finalizing this approach.
Comments containing "TODO" are just bits I have left dangling.

dougbu:
FYI changed base branch to `dev` and added a couple of commits.

dougbu:
Moving forward with something that looks much more like #153.

PR #153, opened 2017-08-27
Add ASP.NET Core WebHooks support (issue #5)

dougbu:
This approach is more ASP.NET Core-native than #152.

Users write controller actions annotated with (say) `[GitHubWebHookAction]` instead of `WebHookHandler`s. Dispatching between multiple handlers must be done manually though we could add abstract controller base classes which dispatch to handlers in the old way.

Those developing new receivers can start from `WebHookReceiver` though the subclasses should also implement `IResourceFilter` or `IAsyncResourceFilter`. This improves modularity and allows base `WebHook...Filter`s for most needs.


src/Microsoft.AspNetCore.WebHooks.Common/Extensions/DictionaryExtensions.cs
```diff
@@ -0,0 +1,80 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.ComponentModel;
+
+namespace System.Collections.Generic
```
davidfowl: Why are these exposed as public API on types we don't own? This should be shared source

src/Microsoft.AspNetCore.WebHooks.Common/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.ComponentModel;
+using System.Linq;
+
+namespace System
```
davidfowl: Same comment as above

src/Microsoft.AspNetCore.WebHooks.Common/Extensions/DictionaryExtensions.cs
dougbu: Absolutely. This was straight out of `Microsoft.AspNet.WebHooks.Common` and I didn't plan to ship with the extension methods in place.

Could skip creating a shared-source package and make these non-extension `static` methods in a `Microsoft.AspNetCore.WebHooks` namespace. The methods are used in various receivers but I'm not sure user code (controller actions in the new world) need them.

Havret:
What is the status of this PR? Can I safely try to use it?

dougbu:
@Havret our team feels the general approach is sound but we have not yet reviewed the details of this PR. Feel free to test it out but I would not recommend call it a perfect indicator of the APIs we'll ship. Put another way, treat `dougbu/add.core.II` as a proposed feature branch with more changes expected.

Note the Core Trello and WordPress receivers are currently almost-empty projects. Core Custom, Generic, Instagram, MyGet, VSTS, and Zendesk receivers do not yet exist; neither does anything on the sender side (no Microsoft.AspNetCore.WebHooks.Custom&hellip;).

Please let us know if you experience any problems!

----

For those of you used to the WebHooks 1.0.0 approach:
- Receiver names shouldn’t have changed
- Configuration settings names and formats shouldn’t have changed though web.config isn’t relevant anymore
  - best set as Azure application settings, environment variables, or in appsettings.json
- Handlers are now ASP.NET Core MVC action methods with `[GitHubWebHook]` attributes (or whichever receiver you choose)
  - check the attribute documentation comments for supported action parameters; names are important here because binding is convention-based
 - A number of receivers do not support binding event names but `KuduConstants` (or whichever) and `IFormCollection`, `JObject`, and  `XElement` action parameters make it pretty each to grab the information from request bodies
- Receivers are now a bundle of metadata-driven MVC application models, constraints, and filters that are (mostly) enabled globally but no-op for receivers that don’t use them
  - configure a receiver in your `Startup.ConfigureServices(…)` method with `services.AddMvcCore().AddAzureAlertWebHooks();` (or whichever)

Havret:
I am trying to build custom ASP.NET Core Receiver based on dougbu/add.core.II. I would be great if you provide any working samples. :)

dougbu:
@Havret I'm not sure what you mean. The branch contains a fair number of different Core receivers as well as sample ASP.NET Core MVC applications using them. Could you clarify your question?

Havret:
@dougbu Sorry, I've got it.

dougbu:
Rebased on `dougbu/stripe.event.data.154` and removed changes covered in #161. Now includes Stripe support in Core receiver and command-line builds are clean.

dougbu:
:up::date: to reduce number of commits

src/Microsoft.AspNetCore.WebHooks.Common/Extensions/WebHookHttpRequestExtensions.cs
```diff
@@ -0,0 +1,100 @@
...
+            if (type.Equals("application", StringComparison.OrdinalIgnoreCase) ||
+                type.Equals("text", StringComparison.OrdinalIgnoreCase))
+            {
+                return contentType.SubType.EndsWith("+xml", StringComparison.OrdinalIgnoreCase);
+            }
+
+            return false;
+        }
+
+        // ??? Should ASP.NET Core expose a similar HttpRequest, HttpContext or ConnectionInfo (extension) method?
```
rynowak: >// ??? Should ASP.NET Core expose a similar HttpRequest, HttpContext or ConnectionInfo (extension) method? [](start = 8, length = 106)

???

src/Microsoft.AspNetCore.WebHooks.Receivers/GeneralWebHookAttribute.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Microsoft.AspNetCore.WebHooks
+{
+    /// <summary>
+    /// An <see cref="Attribute"/> indicating the associated action is a WebHook endpoint for all enabled
+    /// receivers. Specifies the expected <see cref="BodyType"/>, optional <see cref="EventName"/>, and optional
+    /// <see cref="WebHookAttribute.Id"/>. Also adds a <see cref="Filters.WebHookReceiverExistsFilter"/> for
+    /// the action.
+    /// </summary>
+    public class GeneralWebHookAttribute : WebHookAttribute, IWebHookRequestMetadata, IWebHookEventSelectorMetadata
```
rynowak: >GeneralWebHookAttribute [](start = 17, length = 23)

This is a cool idea, might want to call it `GenericWebHookAttribute`

src/Microsoft.AspNetCore.WebHooks.Receivers/Microsoft.AspNetCore.WebHooks.Receivers.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <Description>ASP.NET Core WebHooks abstractions and infrastructure components for receivers. Contains constraints, filters, metadata, and the application models to apply them.</Description>
+    <PackageTags>aspnetcore;webhook;receiver</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="2.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Formatters.Json" Version="2.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Formatters.Xml" Version="2.0.0" />
```
rynowak: ><PackageReference Include="Microsoft.AspNetCore.Mvc.Formatters.Xml" Version="2.0.0" /> [](start = 4, length = 86)

What's the piece that requires the JSON and XML formatters specifically?

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookAttribute.cs
```diff
@@ -0,0 +1,126 @@
...
+    // ??? Microsoft.AspNetCore.Mvc.ViewFeatures reference.
+    // ??? Should this also implement IDisableCorsAttribute to disable CORS? Requires a Microsoft.AspNetCore.Cors
+    // ??? reference.
+    /// <summary>
+    /// Base for <see cref="Attribute"/>s indicating the associated action is a WebHook endpoint. Specifies the
+    /// required (in most cases) <see cref="ReceiverName"/> and optional <see cref="Id"/>. Also adds a
+    /// <see cref="WebHookReceiverExistsFilter"/> for the action.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
+    public abstract class WebHookAttribute : Attribute, IAllowAnonymous, IFilterFactory
```
rynowak: >IAllowAnonymous [](start = 56, length = 15)

Are webhooks **ALWAYS** anonymous?

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookAttribute.cs
```diff
@@ -0,0 +1,126 @@
...
+                throw new ArgumentException(Resources.General_ArgumentCannotBeNullOrEmpty, nameof(receiverName));
+            }
+
+            ReceiverName = receiverName;
+        }
+
+        /// <summary>
+        /// Gets the name of an available <see cref="IWebHookReceiver"/>.
+        /// </summary>
+        public string ReceiverName { get; }
```
rynowak: >public string ReceiverName { get; } [](start = 8, length = 35)

Is this always required? A subclass that uses the default constructor can't set it

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookOptions.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// <summary>
+        /// Gets the list of action parameter <see cref="Type"/>s the
+        /// <see cref="ModelBinding.WebHookHttpContextModelBinder"/> can be associated with. Parameters assignable to
+        /// the given types may be model bound from <see cref="HttpContext.Items"/> entries.
+        /// </summary>
+        /// <value>
+        /// Default collection includes <see cref="IFormCollection"/>, <see cref="JArray"/>, <see cref="JObject"/>, and
+        /// <see cref="XElement"/>.
+        /// </value>
+        public IList<Type> HttpContextItemsTypes { get; } = new List<Type>
```
rynowak: >HttpContextItemsTypes [](start = 27, length = 21)

I don't super know what this is about... I feel like there's a feature here that needs a better explanation

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookReceiverConfig.cs
```diff
@@ -0,0 +1,173 @@
...
+{
+    /// <summary>
+    /// Provides a default <see cref="IWebHookReceiverConfig"/> implementation which manages
+    /// <see cref="IWebHookReceiver"/> configuration using application settings. The name of the application setting is
+    /// '<c>MS_WebHookReceiverSecret_&lt;name&gt;</c>' where '<c>name</c>' is the name of the receiver, for example
+    /// <c>github</c>. The value is a comma-separated list of secrets, using an ID to differentiate between them. For
+    /// example, '<c>secret0, id1=secret1, id2=secret2</c>'. The corresponding WebHook URI is of the form
+    /// '<c>https://&lt;host&gt;/api/webhooks/incoming/custom/{id}</c>'.
+    /// </summary>
+    public class WebHookReceiverConfig : IWebHookReceiverConfig
```
rynowak: >WebHookReceiverConfig [](start = 17, length = 21)

What isn't this just `IConfiguration`. It sounds like there's a feature here that needs to be thought out

src/Microsoft.AspNetCore.WebHooks.Receivers/ApplicationModels/WebHookMetadataProvider.cs
```diff
@@ -0,0 +1,270 @@
...
+        /// </summary>
+        /// <param name="metadata">The collection of <see cref="IWebHookMetadata"/> services.</param>
+        /// <param name="loggerFactory">The <see cref="ILoggerFactory"/>.</param>
+        public WebHookMetadataProvider(
+            IEnumerable<IWebHookMetadata> metadata,
+            ILoggerFactory loggerFactory)
+        {
+            _bindingMetadata = metadata.OfType<IWebHookBindingMetadata>().ToArray();
+            _eventMetadata = metadata.OfType<IWebHookEventMetadata>().ToArray();
+            _requestMetadata = metadata.OfType<IWebHookRequestMetadataService>().ToArray();
```
rynowak: This seems like it's doing an end-run around DI? Why aren't these things just registered in DI with the approriate types?

src/Microsoft.AspNetCore.WebHooks.Receivers/Microsoft.AspNetCore.WebHooks.Receivers.csproj
dougbu: This is really about project file duplication. Most of the receiver-specific projects call `AddJsonFormatters()`, `AddXmlSerializerFormatters()` or both. But, if you prefer, I could move the dependency into those projects. LMK

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookAttribute.cs
dougbu: Yup

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookAttribute.cs
dougbu: Yes, a receiver-specific attribute needs a receiver-specific non-`null` name. `WebHookMetadataProvider` checks this property to make sure it's not `null` except in an attribute that also implements `IWebHookRequestMetadata`. But, there isn't a check to make sure the attribute is `[GeneralWebHook]`.

Easiest thing to do may be to make the default constructor above `internal`. Thoughts?

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookOptions.cs
dougbu: This is about resource filters that call input formatters. When they're done they place the deserialized object into type-named locations in `HttpContext.Items` for use in the `WebHookHttpContextModelBinder`. The mechanism overall exists to reduce the number of times the request body is parsed. This exists to avoid leaking random `HttpContext.Items` entries into the actions and to ensure `WebHookModelBindingProvider` configures the `WebHookHttpContextModelBinder` correctly.

What of that information is needed in the doc comments? Or, are you suggesting we do away with either this knob or the overall "read once" mechanism?

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookReceiverConfig.cs
dougbu: The receivers do contain direct references to the `IConfiguration`. But, this is about specially-formatted configuration entries for secret keys. This uses the same key formats as the Microsoft.AspNet.WebHooks version.

src/Microsoft.AspNetCore.WebHooks.Receivers/ApplicationModels/WebHookMetadataProvider.cs
dougbu: Done this way for flexibility. All of the metadata interfaces inherit from `IWebHookMetadata` and that's how the implementations should be registered. This mechanism allows specific receivers to register a single metadata object once or multiple per-interface instances however much they like.

If I instead had the receivers register using the specific interfaces, could still have one object per receiver that does the implementation. But, the registration would get messy and multiple classes would be aware of the chosen implementations -- the implementation and the extension methods doing registration.

src/Microsoft.AspNetCore.WebHooks.Common/Extensions/WebHookHttpRequestExtensions.cs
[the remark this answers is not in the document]
rynowak: Open a bug :+1: - I don't think this is the right place for this. If we need this code we should inline it or make it internal

---
In reply to: [145833956](https://github.com/aspnet/WebHooks/pull/153#discussion_r145833956) [](ancestors = 145833956)

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookConstants.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <see cref="IWebHookReceiverConfig"/> implementation
+        /// </para>
+        /// </summary>
+        /// <remarks>
+        /// Most, if not all, receiver configurations include
+        /// <see cref="Filters.WebHookSecurityFilter.EnsureSecureConnection"/> calls. For example,
+        /// <see cref="Filters.WebHookVerifyCodeFilter"/> calls the method if
+        /// <see cref="Metadata.IWebHookSecurityMetadata.VerifyCodeParameter"/> is <see langword="true"/>.
+        /// </remarks>
+        public static string DisableHttpsCheckConfigurationKey => "MS_WebHookDisableHttpsCheck";
```
rynowak: >public static string DisableHttpsCheckConfigurationKey => "MS_WebHookDisableHttpsCheck"; [](start = 8, length = 88)

What are these keys into, this isn't the way our IConfiguration system typically works

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookOptions.cs
rynowak: This seems like an implementation detail that is masquerading as a feature, we should talk more about this

---
In reply to: [145838603](https://github.com/aspnet/WebHooks/pull/153#discussion_r145838603) [](ancestors = 145838603)

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookReceiverConfig.cs
```diff
@@ -0,0 +1,173 @@
...
+            var settings = new Dictionary<string, string>(StringComparer.Ordinal);
+
+            foreach (var setting in configuration.AsEnumerable())
+            {
+                var key = setting.Key;
+                if (key.Length > WebHookConstants.ReceiverConfigurationKeyPrefix.Length &&
+                    key.StartsWith(WebHookConstants.ReceiverConfigurationKeyPrefix, StringComparison.OrdinalIgnoreCase))
+                {
+                    // Extract configuration (again, likely receiver) name
+                    var configurationName = key.Substring(WebHookConstants.ReceiverConfigurationKeyPrefix.Length);
```
rynowak: This is interesting, I think this needs some design because stuff like this is pretty high impact and we want it to be consistent across the stack

JDL440:
Is .NET Core 1.0 a possibility?

dougbu:
@JDL440 this PR targets .NET Core 2.0. What drives your 1.0 request?

JDL440:
My service targets 1.0 currently and I'm investigating add webhook subscriptions.

dougbu:
@JDL440 please file a separate issue in this repo so that we can track the idea of re-targeting the new ASP.NET Core support.

Another issue, perhaps filed at https://github.com/aspnet/Home, may help us understand anything blocking migrating your service to .NET Core 2.0. For example, is the documentation at https://docs.microsoft.com/en-us/aspnet/core/migration/1x-to-2x/ insufficient in some way?

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookReceiverConfig.cs
dougbu: As discussed offline, I'll switch to `IConfiguration` extension methods supporting a consistent hierarchy of configuration sections e.g.
``` json
"WebHooks": {
  "GitHub": {
    "SecretKey": {
      "default": "default secret key",
      "id1": "secret key for id1"
    }
  }
}
```
I don't think I'll make `"default"` a reserved word because nothing breaks when a URL contains that id. Instead `https://{host}/api/webhooks/incoming/github` and `https://{host}/api/webhooks/incoming/github/default` will use the same secret key.

The other option looks like
``` json
"WebHooks": {
  "GitHub": {
    "SecretKey": {
      "": "default secret key",
      "id1": "secret key for id1"
    }
  }
}
```
That is valid JSON but looks very strange. And I'm not sure our configuration system supports empty property names.

dougbu:
5f62336 got this in because it's too big to review. Will address any issues in follow-up PRs.

PR #158, opened 2017-09-22
Fix heading syntax for GitHub flavoured markdown

leggetter:
Headings will now render on GitHub and provide inline anchors.


Eilon:
Thanks!

PR #161, opened 2017-10-02
Suggest changes to two quirky-looking bits

dougbu:
Update two quirky-looking bits of code
- use SOAP fault when Salesforce handler fails
  - preserve non-200 but successful HTTP status code
- add lots of comments about `IsAssignableFrom(...)` use in `GetDataOrDefault<T>(...)`
  - test this method more thoroughly

nits:
- use new C# features
- remove duplicate test data in `GenericJsonWebHookReceiverTests`
- take VS suggestions, especially in `WebHookHandlerContextExtensionsTests`

### DIFF
--- a/src/Microsoft.AspNet.WebHooks.Receivers.Salesforce/WebHooks/SalesforceSoapWebHookReceiver.cs
+++ b/src/Microsoft.AspNet.WebHooks.Receivers.Salesforce/WebHooks/SalesforceSoapWebHookReceiver.cs
@@ -93,12 +93,28 @@ namespace Microsoft.AspNet.WebHooks
                 // Call registered handlers
                 var response = await ExecuteWebHookAsync(id, context, request, new string[] { action }, notifications);
 
-                // Add SOAP response if not already present
-                if (response == null || response.Content == null || !response.Content.IsXml())
+                // Add SOAP response content if not already present or isn't XML. Ignore current (e.g. JSON) content.
+                if (response?.Content == null || !response.Content.IsXml())
                 {
-                    var success = ReadResource("Microsoft.AspNet.WebHooks.Messages.NotificationResponse.xml");
-                    response = GetXmlResponse(request, HttpStatusCode.OK, success);
+                    // Ignore redirects because SOAP 1.1 doesn't mention them and they're corner cases in SOAP.
+                    var statusCode = response?.StatusCode ?? HttpStatusCode.OK;
+                    if (statusCode >= (HttpStatusCode)200 && statusCode < (HttpStatusCode)300)
+                    {
+                        var success = ReadResource("Microsoft.AspNet.WebHooks.Messages.NotificationResponse.xml");
+                        response = GetXmlResponse(request, statusCode, success);
+                    }
+                    else
+                    {
+                        // Move failure information into a SOAP fault response. Fault contains code soapenv:Client and
+                        // that must be transmitted with HTTP status 400, Bad Request according to SOAP 1.2 (mixing
+                        // that sensible choice into this SOAP 1.1 implementation).
+                        var resource = ReadResource("Microsoft.AspNet.WebHooks.Messages.FaultResponse.xml");
+                        var faultString = $"Handler's status: {statusCode}, '{response.ReasonPhrase}'.";
+                        var failure = string.Format(CultureInfo.InvariantCulture, resource, faultString);
+                        response = GetXmlResponse(request, HttpStatusCode.BadRequest, failure);
+                    }
                 }
+
                 return response;
             }
             else
--- a/src/Microsoft.AspNet.WebHooks.Receivers/Extensions/WebHookHandlerContextExtensions.cs
+++ b/src/Microsoft.AspNet.WebHooks.Receivers/Extensions/WebHookHandlerContextExtensions.cs
@@ -30,16 +30,34 @@ namespace Microsoft.AspNet.WebHooks
                 return default(T);
             }
 
-            if (context.Data is JToken && !typeof(JToken).IsAssignableFrom(typeof(T)))
+            // IsAssignableFrom(...) check looks odd because return statement effectively assigns from Data to T.
+            // However, this check avoids useless "conversions" e.g. from a JObject to a JObject without attempting
+            // impossible conversions e.g. from a JObject to a JArray. Json.NET does not support semantically-invalid
+            // conversions between JToken subtypes.
+            //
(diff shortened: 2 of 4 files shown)


dougbu:
Rebased on `dougbu/infrastructure.updates` and changed questions into proposed answers. Bit easier to review with `files?w=1` in the URL.

@HenrikFrystykNielsen I would appreciate your thoughts on the `SalesforceSoapWebHookReceiver` and `WebHookHandlerContextExtensions` changes. Am I missing something? And, why aren't the unit tests affected? (Should I add more tests?)

dougbu:
Reopening to see if AppVeyor config is working

dougbu:
@HenrikFrystykNielsen there's now nothing in this PR except my suggestions (and a few trailing whitespace removals). Please have a look&hellip; I am not confident these changes are correct for the existing receivers.

src/Microsoft.AspNet.WebHooks.Receivers.Salesforce/WebHooks/SalesforceSoapWebHookReceiver.cs
```diff
@@ -93,12 +93,17 @@ public override string Name
...
                 if (response == null || response.Content == null || !response.Content.IsXml())
                 {
-                    var success = ReadResource("Microsoft.AspNet.WebHooks.Messages.NotificationResponse.xml");
-                    response = GetXmlResponse(request, HttpStatusCode.OK, success);
+                    var statusCode = response?.StatusCode ?? HttpStatusCode.OK;
+                    if (statusCode >= (HttpStatusCode)200 && statusCode < (HttpStatusCode)300)
+                    {
+                        var success = ReadResource("Microsoft.AspNet.WebHooks.Messages.NotificationResponse.xml");
+                        response = GetXmlResponse(request, statusCode, success);
+                    }
```
dougbu: Add `else` to handle the failure case w/ the SOAP error response.

src/Microsoft.AspNet.WebHooks.Receivers/Extensions/WebHookHandlerContextExtensions.cs
```diff
@@ -30,8 +30,10 @@ public static T GetDataOrDefault<T>(this WebHookHandlerContext context)
                 return default(T);
             }
 
-            if (context.Data is JToken && !typeof(JToken).IsAssignableFrom(typeof(T)))
+            if (context.Data is JToken && !typeof(T).IsAssignableFrom(context.Data.GetType()))
```
dougbu: Add tests to confirm potential failure cases e.g. `T` is `JArray` but the current object is `JObject`.

dougbu: Testing showed this change only affects corner cases. It's less efficient in some of those cases e.g. where `Data` is a `JObject` and `T` is `JArray`. Will attempt the impossible conversion and catch the thrown `Exception` rather than simply returning `null`. Don't see a compelling reason to do more than add comments and tests.

dougbu:
With thanks for his offline comments, removing @HenrikFrystykNielsen's review request.

dougbu:
:up::date: and ready for @rynowak

src/Microsoft.AspNet.WebHooks.Receivers.Salesforce/WebHooks/SalesforceSoapWebHookReceiver.cs
```diff
@@ -93,12 +93,28 @@ public override string Name
...
+                    else
+                    {
+                        // Move failure information into a SOAP fault response. Fault contains code soapenv:Client and
+                        // that must be transmitted with HTTP status 400, Bad Request according to SOAP 1.2 (mixing
+                        // that sensible choice into this SOAP 1.1 implementation).
+                        var resource = ReadResource("Microsoft.AspNet.WebHooks.Messages.FaultResponse.xml");
+                        var faultString = $"Handler's status: {statusCode}, '{response.ReasonPhrase}'.";
+                        var failure = string.Format(CultureInfo.InvariantCulture, resource, faultString);
```
rynowak: Should this be in a resource since it's going out of the wire?

src/Microsoft.AspNet.WebHooks.Receivers/Extensions/WebHookHandlerContextExtensions.cs
```diff
@@ -30,16 +30,34 @@ public static T GetDataOrDefault<T>(this WebHookHandlerContext context)
...
+            // !typeof(T).IsAssignableFrom(context.Data.GetType()) may be more precise but is less efficient. That
+            // check would not change the (null) outcome in the JObject to JArray case, just adds a first-chance
+            // Exception (because the code would attempt the impossible conversion). About the only new cases it
+            // handles with a cast instead of ToObject<T>() are extreme corner cases such as when T is
+            // IDictionary<string, JToken> or another interface the current Data (e.g. JObject) may implement. Even
+            // then, Json.NET can usually perform an explicit conversion.
+            if (context.Data is JToken token && !typeof(JToken).IsAssignableFrom(typeof(T)))
             {
+                // context.Data is a JToken and the caller isn't just asking for (say) the current JObject.
+                // Do explicit conversion.
```
rynowak: So there's this huge comment here explaining the design tradeoffs but what does this code **mean**? Does it basically say that this block handles the case where T is one of the JLinq types?

dougbu: The block handles the case where `Data` is one of the JLinq types and `T` is *not* e.g. `Data` is a `JObject` and `T` is something like `AzureAlertNotification` (one of the strong types for request content). I'll add this info before the design "discussion".

src/Microsoft.AspNet.WebHooks.Receivers.Salesforce/WebHooks/SalesforceSoapWebHookReceiver.cs
dougbu: Which "this"? If you mean the `faultString`, perhaps. I'll also check other places we use the `FaultResponse.xml` resource.

rynowak: Yeah I just wondered. Even if we don't do loc it's nice to have a separation between strings that a user sees and strings that are just used as IDs.

dougbu:
78491ef819

PR #169, opened 2017-10-09
Add `$(TestInParallel)`

dougbu:
- allows user to e.g. build in parallel but test assemblies serially
- default remains doing everything in parallel


dougbu:
a735611

PR #172, opened 2017-10-30
Address some PR comments from #153

dougbu:
- more for #5
- grab some fixes from aspnet/AspNetWebStack repo
- revamp `IConfiguration` use
  - remove `IWebHookReceiverConfig`, its implementation and extension methods
  - make methods synchronous if they were `async` only for configuration lookups
  - remove configuration value caching in `PusherVerifySignatureFilter`
- remove `HttpRequest.IsLocal()` extension method
  - instead check for development environment
- move JSON and XML dependencies into specific receiver projects
- make default `WebHookAttribute` constructor `internal` and move its doc comments to `GeneralWebHookAttribute`
- add `WebHookConstants.EventKeyNames`
  - add `RouteData.TryGetEventName(...)` extension method; `WebHookPingResponseFilter` does not need full array

Note this PR does not close questions @rynowak and I had left unfinished in #153.


dougbu:
Please let me know if I should split the actual PR comments commit (the middle one) into a separate PR. Would be good to get this repo building on our CI soon.

dougbu:
Reopening to test out AppVeyor

src/Microsoft.AspNetCore.WebHooks.Common/Extensions/WebHookConfigurationRootExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+using System;
+using System.ComponentModel;
+
+namespace Microsoft.Extensions.Configuration
+{
+    /// <summary>
+    /// Extension methods for <see cref="IConfigurationRoot"/>.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static class WebHookConfigurationRootExtensions
```
rynowak: >WebHookConfigurationRootExtensions [](start = 24, length = 34)

We should not be adding things like, and certainly not in the webhooks repo

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookAttribute.cs
```diff
@@ -24,28 +24,12 @@ public abstract class WebHookAttribute : Attribute, IAllowAnonymous, IFilterFact
...
-        /// The default route <see cref="Mvc.Routing.IRouteTemplateProvider.Name"/> is <see langword="null"/>.
-        /// </para>
-        /// </summary>
-        protected WebHookAttribute()
+        internal WebHookAttribute()
         {
         }
 
+        // TODO: Move attribute constructors' comments, especially the recommended action signatures, to class level.
+        // TODO:  Do the same for all subclasses. As-is important information is hard to find.
```
rynowak: Can we use issues for these things instead of comments?

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookConstants.cs
```diff
@@ -32,43 +34,61 @@ public static class WebHookConstants
...
         /// </summary>
         /// <seealso cref="Metadata.IWebHookEventMetadata"/>
         public static string EventKeyName => "event";
 
         /// <summary>
+        /// Gets the names of the <see cref="AspNetCore.Routing.RouteValueDictionary"/> entries containing event names
+        /// for the current request when there are more than one event name.
+        /// </summary>
+        /// <seealso cref="Metadata.IWebHookEventMetadata"/>
+        public static string[] EventKeyNames { get; } = Enumerable.Range(0, 100)
```
rynowak: >EventKeyNames [](start = 31, length = 13)

Why do we need this as a public constant? why would someone developing a receiver care

src/Microsoft.AspNetCore.WebHooks.Common/Extensions/WebHookConfigurationRootExtensions.cs
dougbu: The plan (#174) is to merge Common into the Receivers package. I could move this up into `WebHookReceiverConfigurationExtensions` to be proactive if you feel strongly about it.

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookConstants.cs
dougbu: Because otherwise every request would create at least two `string`s per entry in the array. And requests an contain numerous notifications / event names.

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookAttribute.cs
dougbu: How about a blanket "Address Core comments and TODOs" issue? There are a fair number of dangling `// TODO` and `// ???` comments. I'd rather do a (separate) single PR to do the work than create the issues and do an interim PR to remove the comments.

I'm just moving this particular TODO from one place to another.

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookConstants.cs
dougbu: Also, users can write their own receivers to special-case event names in some way.

The alternative is to add a `SetEventNames(...)` extension method to `WebHookRouteDataExtensions`. Does that feel better to you?

src/Microsoft.AspNetCore.WebHooks.Receivers/Extensions/WebHookReceiverConfigurationRootExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// <paramref name="sectionKey"/> and <see cref="WebHookConstants.SecretKeyConfigurationKeySectionKey"/>) of
+        /// the <see cref="IConfigurationSection"/> to return. This allows an <see cref="IWebHookReceiver"/> to support
+        /// multiple WebHook senders with individual
+        /// configurations.
+        /// </param>
+        /// <returns>
+        /// The secret key <see cref="IConfigurationSection"/> with the given <paramref name="sectionKey"/> and
+        /// <paramref name="id"/>. <see langword="null"/> if the <see cref="IConfigurationSection"/> does not exist.
+        /// </returns>
+        public static IConfigurationSection GetSecretKeys(
```
rynowak: >GetSecretKeys [](start = 44, length = 13)

Should we put webhooks in these names like `GetWebHooksSecretKeys`? This makes it sound like this functionality is for generic secrets - this is really focused on how webooks expects its configuration to work

src/Microsoft.AspNetCore.WebHooks.Receivers/Extensions/WebHookReceiverConfigurationRootExtensions.cs
```diff
@@ -0,0 +1,107 @@
...
+                WebHookConstants.ReceiverConfigurationSectionKey,
+                sectionKey,
+                WebHookConstants.SecretKeyConfigurationKeySectionKey,
+                id);
+
+            var section = configuration.GetSection(key);
+            if (!section.Exists())
+            {
+                return null;
+            }
```
rynowak: It really seems like we should follow the conventions of IConfiguration here. If IConfiguration uses the null-object-pattern then we should as well.

rynowak:
> init:

Per your request, pausing my reivew until you've have a chance to update :)

---
Refers to: .appveyor.yml:1 in 3a50b95. [](commit_id = 3a50b95d546aafc36066c290fed9bed66670b571, deletion_comment = False)

dougbu:
:up::date:

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookAttribute.cs
rynowak: I don't understand what we gain by adding MORE todos to the code.

We use issues to track work. This PR adds todos. 

---
In reply to: [150085349](https://github.com/aspnet/WebHooks/pull/172#discussion_r150085349) [](ancestors = 150085349)

src/Microsoft.AspNetCore.WebHooks.Receivers/Extensions/WebHookReceiverConfigurationExtensions.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        /// <summary>
+        /// Returns <see langword="true"/> if the configuration value with given <paramref name="key"/> is set to
+        /// 'true'; otherwise <see langword="false"/>.
+        /// </summary>
+        /// <param name="configuration">The <see cref="IConfiguration"/>.</param>
+        /// <param name="key">The key of the configuration value to evaluate.</param>
+        /// <returns><see langword="true"/> if the value is set to 'true'; otherwise <see langword="false"/>.</returns>
+        public static bool IsTrue(this IConfiguration configuration, string key)
```
rynowak: >IsTrue [](start = 27, length = 6)

We should not have general purpose extension methods like this coming from this repo. If there is a need for something like this, open an issue on the configuration repo and drive it there.

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookSecurityFilter.cs
```diff
@@ -89,41 +103,39 @@ protected WebHookSecurityFilter(ILoggerFactory loggerFactory, IWebHookReceiverCo
...
         // ??? Why is this called so rarely? See Dropbox, GitHub and Pusher filters and corresponding old receivers.
         /// <summary>
         /// Some WebHooks rely on HTTPS for sending WebHook requests in a secure manner. A
         /// <see cref="WebHookSecurityFilter"/> subclass can call this method to ensure that the incoming WebHook
         /// request is using HTTPS. If the request is not using HTTPS an error will be generated and the request will
         /// not be further processed.
         /// </summary>
-        /// <remarks>This method does allow local HTTP requests using <c>localhost</c>.</remarks>
+        /// <remarks>
+        /// This method allows non-HTTP requests while the application is in development or if the
```
rynowak: >This method allows non-HTTP  [](start = 12, length = 28)

non-HTTPS I think

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookSecurityFilter.cs
```diff
@@ -149,80 +161,124 @@ protected virtual IActionResult EnsureSecureConnection(string receiverName, Http
...
+            string sectionKey,
             RouteData routeData,
-            string configurationName,
             int minLength,
             int maxLength)
         {
-            if (request == null)
+            // Look up configuration for this receiver and instance.
+            var secrets = GetSecretKeys(sectionKey, routeData);
+            if (secrets == null)
```
rynowak: >null [](start = 27, length = 4)

I think should should be a .Exists() check not a null check

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookConstants.cs
rynowak: Yes, the extension method you added for this seems fine.

---
In reply to: [150086244](https://github.com/aspnet/WebHooks/pull/172#discussion_r150086244) [](ancestors = 150086244)

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookSecurityFilter.cs
rynowak: Are there more cases like this - where we're null checking the result of getting a config section?

---
In reply to: [150144519](https://github.com/aspnet/WebHooks/pull/172#discussion_r150144519) [](ancestors = 150144519)

src/Microsoft.AspNetCore.WebHooks.Receivers/Extensions/WebHookReceiverConfigurationExtensions.cs
dougbu: How will a new method in 2.1.0 help the current WebHooks implementation?

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookAttribute.cs
dougbu: I'm confused. I mentioned
> I'm just moving this particular TODO from one place to another.

And you replied
> This PR adds todos.

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookSecurityFilter.cs
dougbu: Good catch. There were two places where I was still `null`-checking `IConfigurationSection`s.

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookSecurityFilter.cs
dougbu: Went with "HTTP".

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookAttribute.cs
dougbu: See #178

src/Microsoft.AspNetCore.WebHooks.Receivers/Extensions/WebHookReceiverConfigurationExtensions.cs
dougbu: Will cover this as part of #178.

dougbu:
Minor :up::date: is just FYI.

dougbu:
@rynowak please dummy-check the few changes in 1f36dd6e1e. `StripeVerifyNotificationIdFilter` is about the only class with a significant revision.

dougbu:
8f0a19bfb9

PR #176, opened 2017-11-09
Quick fix: Add .gitattributes and normalize line endings

dougbu:
- followed steps from https://stackoverflow.com/a/1511273
- mostly affected the Custom senders and their tests


src/Microsoft.AspNet.WebHooks.Custom.AzureStorage/WebHooks/AzureWebHookDequeueManager.cs
```diff
@@ -276,9 +276,9 @@ public QueuedSender(AzureWebHookDequeueManager parent, ILogger logger)
                             deleteMessages.Add(message);
                         }
                     }
-                }
-
-                // Loop through the responses to see which messages should be deleted from the queue based on the response statuses.
+                }
+
+                // Loop through the responses to see which messages should be deleted from the queue based on the response statuses.
```
dougbu: This is the opposite of the change in commit c3890ceb0b which is part of #161. VS fixed the mixed line endings but landed with Unix endings even on Windows.

The other files in this PR just had Unix line endings even on Windows with `core.autocrlf` set to `true`.

dougbu:
@rynowak please look at this before #172. I want to get it in and rebase other PRs. One of my local repos is messed up without these changes.

https://github.com/aspnet/WebHooks/pull/176/files?w=1 gives a nice view of how little I changed.

dougbu:
be8571953a

PR #184, opened 2017-11-11
Quick fixes

dougbu:
- remove `WebHookMultipleEventMapperConstraint` and `WebHookSingleEventMapperConstraint`
- use arrays (not lists) when only `IReadOnlyList<T>` is needed
- don't use `GetType()` when creating a `private` `ILogger` implementation


dougbu:
Ping @rynowak (with thanks for #186 review!)

dougbu:
88a56f7749

Issue #191, opened 2017-11-24
Custom configuration for SqlStorage


Marusyk:
To better integration with existing project it would be great to have a more custom configurations. 

For now to store WebHook Registrations in SQL I need connection string in Web.config with name **MS_SqlStoreConnectionString**

```
 <add name="MS_SqlStoreConnectionString" connectionString="Data Source=localhost\SQLEXPRESS;Initial Catalog=Platform.Data;Integrated Security=True" providerName="System.Data.SqlClient" />
```

But I want to store WebHook Registrations in the same database that my application uses, so I already have connection string like:
```
 <add name="Platfom" connectionString="Data Source=localhost\SQLEXPRESS;Initial Catalog=Platform.Data;Integrated Security=True" providerName="System.Data.SqlClient" />
```
In this case I need to add two almost the same connection strings (only the name will be different)

Also, according to our sql conventions we shouldn't create a new Schema. 
I think, it would be nice to have custom configuration for that parameters (maybe the name of table too). And using it like:
```
config.InitializeCustomWebHooksSqlStorage(encryptData: true, connectionStringName: "Platform", schema: "plt", tableName:"CustomWebHooks");
```

What do you think about that? If ok, I could prepare a PR.

dougbu:
The value of adding a new configuration value that only names another configuration value is not clear to me. Do you have any other use cases beyond not copying-and-pasting reused connection strings?

Marusyk:
Yes, I need other schema name. Also connection string should be secure, and it is simpler to manage one string

Marusyk:
@dougbu I'm sorry but I need to know what do you think about it. We need this feature and we with customers still waiting for your response. Can it be added to the next version (maybe with different implementation) or you can advise some other solution how to achieve our requirements? 
Thanks you in advance

dougbu:
We've discussed and I will work with you on #192.

dougbu:
6c173a343f

dougbu:
Thanks very much @Marusyk ❕

Marusyk:
Thank you for your time

Marusyk:
Hello, are there any news about the future release of nuget?

PR #192, opened 2017-11-24
Custom configuration for SqlStorage

Marusyk:
According to the issue #191
Please review.

### DIFF
--- a/src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/Extensions/HttpConfigurationExtensions.cs
+++ b/src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/Extensions/HttpConfigurationExtensions.cs
@@ -36,7 +36,7 @@ namespace System.Web.Http
         /// <param name="encryptData">Indicates whether the data should be encrypted using <see cref="IDataProtector"/> while persisted.</param>
         public static void InitializeCustomWebHooksSqlStorage(this HttpConfiguration config, bool encryptData)
         {
-            InitializeCustomWebHooksSqlStorage(config, encryptData, null);
+            InitializeCustomWebHooksSqlStorage(config, encryptData, nameOrConnectionString: null);
         }
 
         /// <summary>
--- a/src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/Extensions/HttpConfigurationExtensions.cs
+++ b/src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/Extensions/HttpConfigurationExtensions.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.ComponentModel;
 using System.Data.Entity;
 using Microsoft.AspNet.WebHooks;
@@ -35,24 +34,58 @@ namespace System.Web.Http
         /// </summary>
         /// <param name="config">The current <see cref="HttpConfiguration"/>config.</param>
         /// <param name="encryptData">Indicates whether the data should be encrypted using <see cref="IDataProtector"/> while persisted.</param>
-        /// <param name="connectionStringName">Sets custom name of connection string <see cref="WebHookStoreContext"/></param>
-        /// <param name="schemaName">Sets custom name of schema <see cref="WebHookStoreContext"/></param>
-        /// <param name="tableName">Sets custom name of table <see cref="WebHookStoreContext"/></param>
-        public static void InitializeCustomWebHooksSqlStorage(this HttpConfiguration config, bool encryptData, string connectionStringName = "", string schemaName = "", string tableName = "")
+        public static void InitializeCustomWebHooksSqlStorage(this HttpConfiguration config, bool encryptData)
+        {
+            InitializeCustomWebHooksSqlStorage(config, encryptData: encryptData, nameOrConnectionString: "");
+        }
+
+        /// <summary>
+        /// Configures a Microsoft SQL Server Storage implementation of <see cref="IWebHookStore"/>
+        /// which provides a persistent store for registered WebHooks used by the custom WebHooks module.
+        /// </summary>
+        /// <param name="config">The current <see cref="HttpConfiguration"/>config.</param>
+        /// <param name="encryptData">Indicates whether the data should be encrypted using <see cref="IDataProtector"/> while persisted.</param>
+        /// <param name="nameOrConnectionString">The custom connection string or name of the connection string application setting. Used to initialize <see cref="WebHookStoreContext"/>.</param>
+        public static void InitializeCustomWebHooksSqlStorage(this HttpConfiguration config, bool encryptData, string nameOrConnectionString)
+        {
+            if (nameOrConnectionString == null)
+            {
+                throw new ArgumentNullException(nameof(nameOrConnectionString));
+            }
+
+            InitializeCustomWebHooksSqlStorage(config, encryptData: encryptData, nameOrConnectionString: nameOrConnectionString, schemaName: "", tableName: "");
+        }
+
+        /// <summary>
+        /// Configures a Microsoft SQL Server Storage implementation of <see cref="IWebHookStore"/>
+        /// which provides a persistent store for registered WebHooks used by the custom WebHooks module.
+        /// </summary>
+        /// <param name="config">The current <see cref="HttpConfiguration"/>config.</param>
+        /// <param name="encryptData">Indicates whether the data should be encrypted using <see cref="IDataProtector"/> while persisted.</param>
+        /// <param name="nameOrConnectionString">The custom connection string or name of the connection string application setting. Used to initialize <see cref="WebHookStoreContext"/>.</param>
+        /// <param name="schemaName">The custom name of database schema. Used to initialize <see cref="WebHookStoreContext"/>.</param>
+        /// <param name="tableName">The custom name of database table. Used to initialize <see cref="WebHookStoreContext"/>.</param>
+        public static void InitializeCustomWebHooksSqlStorage(this HttpConfiguration config, bool encryptData, string nameOrConnectionString, string schemaName, string tableName)
         {
             if (config == null)
             {
                 throw new ArgumentNullException(nameof(config));
             }
 
+            if (schemaName == null)
+            {
+                throw new ArgumentNullException(nameof(schemaName));
+            }
+
+            if (tableName == null)
+            {
+                throw new ArgumentNullException(nameof(tableName));
+            }
+
             WebHooksConfig.Initialize(config);
 
             ILogger logger = config.DependencyResolver.GetLogger();
             SettingsDictionary settings = config.DependencyResolver.GetSettings();
-
-            WebHookStoreContext.ConnectionStringName = connectionStringName;
-            WebHookStoreContext.SchemaName = schemaName;
-            WebHookStoreContext.TableName = tableName;
 
             // We explicitly set the DB initializer to null to avoid that an existing DB is initialized wrongly.
             Database.SetInitializer<WebHookStoreContext>(null);
@@ -61,12 +94,13 @@ namespace System.Web.Http
             if (encryptData)
             {
                 IDataProtector protector = DataSecurity.GetDataProtector();
-                store = new SqlWebHookStore(settings, protector, logger);
+                store = new SqlWebHookStore(settings, protector, logger, nameOrConnectionString, schemaName, tableName);
             }
             else
             {
-                store = new SqlWebHookStore(settings, logger);
+                store = new SqlWebHookStore(settings, logger, nameOrConnectionString, schemaName, tableName);
             }
+
             CustomServices.SetStore(store);
         }
     }
--- a/src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/Extensions/HttpConfigurationExtensions.cs
+++ b/src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/Extensions/HttpConfigurationExtensions.cs
@@ -36,7 +36,7 @@ namespace System.Web.Http
         /// <param name="encryptData">Indicates whether the data should be encrypted using <see cref="IDataProtector"/> while persisted.</param>
         public static void InitializeCustomWebHooksSqlStorage(this HttpConfiguration config, bool encryptData)
         {
-            InitializeCustomWebHooksSqlStorage(config, encryptData: encryptData, nameOrConnectionString: string.Empty);
+            InitializeCustomWebHooksSqlStorage(config, encryptData, null);
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


Marusyk:
This is only prototype how it can be implemented. It is very important to get your comments.

src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/Extensions/HttpConfigurationExtensions.cs
```diff
@@ -35,7 +35,10 @@ public static void InitializeCustomWebHooksSqlStorage(this HttpConfiguration con
         /// </summary>
         /// <param name="config">The current <see cref="HttpConfiguration"/>config.</param>
         /// <param name="encryptData">Indicates whether the data should be encrypted using <see cref="IDataProtector"/> while persisted.</param>
-        public static void InitializeCustomWebHooksSqlStorage(this HttpConfiguration config, bool encryptData)
+        /// <param name="connectionStringName">Sets custom name of connection string <see cref="WebHookStoreContext"/></param>
+        /// <param name="schemaName">Sets custom name of schema <see cref="WebHookStoreContext"/></param>
+        /// <param name="tableName">Sets custom name of table <see cref="WebHookStoreContext"/></param>
```
dougbu: - The `connectionStringName` parameter is passed directly to the `DbContext` constructor. Therefore, it's actually a `nameOrConnectionString`. Change the parameter and the `WebHookStoreContext` property's names to `nameOrConnectionString` / `NameOrConnectionString`.
- end all descriptions with a period
- `<see cref="..."/>` just adds a link, doesn't add words like "see also"
- better to reference the specific `WebHookStoreContext` properties in these descriptions (if all remain) or add a general `<see cref="WebHookStoreContext"/>` below these `<param .../>` elements.

E.g.
``` c#
<param name="nameOrConnectionString">The custom connection string or name of the connection string application setting. Used to initialize <see cref="WebHookStoreContext.ConnectionStringName"/>.</param>
```

Marusyk: One question, should be `NameOrConnectionString` or `NameOfConnectionString` ?

src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/WebHooks/WebHookStoreContext.cs
```diff
@@ -12,16 +13,71 @@ namespace Microsoft.AspNet.WebHooks
...
-        private const string ConnectionStringNameParameter = "name=" + ConnectionStringName;
+        private static string _connectionStringName = "MS_SqlStoreConnectionString";
+        private static string _schemaName = "WebHooks";
+        private static string _tableName = "WebHooks";
 
         /// <summary>
         /// Initializes a new instance of the <see cref="WebHookStoreContext"/> class.
         /// </summary>
-        public WebHookStoreContext() : base(ConnectionStringNameParameter)
+        public WebHookStoreContext() : base(ConnectionStringName)
```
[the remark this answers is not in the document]
Marusyk: `connectionStringName` checks by `DbContext` constructor. So, we don't need to add this check.

Marusyk:
Thanks for review. I made fixes. Please review again.

src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/WebHooks/WebHookStoreContext.cs
```diff
@@ -12,16 +15,62 @@ namespace Microsoft.AspNet.WebHooks
...
         {
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="WebHookStoreContext"/> class using the given string
+        /// as the name or connection string for the database to which a connection will be made.
+        /// </summary>
+        /// <param name="nameOrConnectionString">Either the database name or a connection string.</param>
+        public WebHookStoreContext(string nameOrConnectionString)
+            : base(GetConnectionStringNameParameter(nameOrConnectionString))
```
dougbu: This ties the `HttpConfigurationExtensions` caller's hands and needs to go away. We specifically want users to be able to pass the connection string itself into `InitializeCustomWebHooksSqlStorage(...)`.

Remove `GetConnectionStringNameParameter(...)`, restore `ConnectionStringNameParameter`, and pass whatever `nameOrConnectionString` value this is given directly to the base `DbContext`.

Marusyk: Then it's not make sense to check `nameOrConnectionString` because it will be done by `DbContext`

src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/Extensions/HttpConfigurationExtensions.cs
```diff
@@ -36,12 +35,53 @@ public static void InitializeCustomWebHooksSqlStorage(this HttpConfiguration con
         /// <param name="config">The current <see cref="HttpConfiguration"/>config.</param>
         /// <param name="encryptData">Indicates whether the data should be encrypted using <see cref="IDataProtector"/> while persisted.</param>
         public static void InitializeCustomWebHooksSqlStorage(this HttpConfiguration config, bool encryptData)
+        {
+            InitializeCustomWebHooksSqlStorage(config, encryptData: encryptData, nameOrConnectionString: "");
```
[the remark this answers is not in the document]
Marusyk: > reduce the number of overloads
What do you mean? Remove this overload `void InitializeCustomWebHooksSqlStorage(this HttpConfiguration config, bool encryptData, string nameOrConnectionString)` and use only the longest with default params ?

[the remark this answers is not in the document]
Marusyk: > we don't need quite so many overloads 

I created it according to previous comments. Please suggest what overloads can I remove

src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/WebHooks/SqlWebHookStore.cs
```diff
@@ -33,6 +37,54 @@ public SqlWebHookStore(SettingsDictionary settings, ILogger logger)
             CheckSqlStorageConnectionString(settings);
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SqlWebHookStore"/> class with the given <paramref name="settings"/>,
+        /// <paramref name="logger"/> and <paramref name="nameOrConnectionString"/>. 
+        /// Using this constructor, the data will not be encrypted while persisted to the database.
+        /// </summary>
+        public SqlWebHookStore(SettingsDictionary settings, ILogger logger, string nameOrConnectionString)
```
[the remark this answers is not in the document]
Marusyk: If we remove this overload `public SqlWebHookStore(SettingsDictionary settings, IDataProtector protector, ILogger logger, string nameOrConnectionString`  then consumers should always pass table and schema name or nulls for that params even if they do not need it  Is it ok?

Marusyk:
I'm not sure that I properly understood all your comments, but I prepared commit and leave some questions. Thank you in advance

dougbu:
I've updated `dev` to fix the AppVeyor failure. Please rebase your fork when you update this PR.

Marusyk:
Thank you very much for your time and for great review. I made some fixes. Please let me know where we can improve that code. 

I will try to write test by myself during this week. I don't have enough experience with tests but it's really interesting.

src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/Extensions/HttpConfigurationExtensions.cs
```diff
@@ -36,6 +35,37 @@ public static void InitializeCustomWebHooksSqlStorage(this HttpConfiguration con
         /// <param name="config">The current <see cref="HttpConfiguration"/>config.</param>
         /// <param name="encryptData">Indicates whether the data should be encrypted using <see cref="IDataProtector"/> while persisted.</param>
         public static void InitializeCustomWebHooksSqlStorage(this HttpConfiguration config, bool encryptData)
+        {
+            InitializeCustomWebHooksSqlStorage(config, encryptData, null);
```
dougbu: Name the `null` argument value.

dougbu:
> I will try to write test by myself during this week. I don't have enough experience with tests but it's really interesting.

Sounds fine. We're not in a big rush.

Marusyk:
Hello, I'm trying to add unit test to `HttpConfigurationExtensionsTests` and `SqlWebHookStoreTests` and I'm a little bit confused. 

Does it make sense to add test on each overload of `InitializeCustomWebHooksSqlStorage`?
The same for `CreateStore_Succeeds()` from `SqlWebHookStoreTests`.

What about `IWebHookStore CreateStore()`, should I pass custom connectionString, tablename and schema and check if database create table with passed tablename?

Thank you in advance

dougbu:
@Marusyk thanks for your attempt. I'll have a look at the tests later today. Hopefully, I can see why the AppVeyor build is failing.

> Does it make sense to add test on each overload of `InitializeCustomWebHooksSqlStorage`?

In this case, probably. The usual idea is to test each unique code path and, in this case, that means passing `null` and non-`null` values. The shorter overloads are the easy way to pass `null` values to the longest overload.

> The same for `CreateStore_Succeeds()` from `SqlWebHookStoreTests`.

Since `CreateStore(...)` does little more than call the `SqlWebHookStore`, your added test is enough.

> I think, I have to revert that last commit

Why? The CI failures can likely be fixed.

src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/WebHooks/SqlWebHookStore.cs
```diff
@@ -66,21 +111,54 @@ public static IWebHookStore CreateStore(ILogger logger)
...
+
+        /// <summary>
+        /// Provides a static method for creating a standalone <see cref="SqlWebHookStore"/> instance.
+        /// </summary>
+        /// <param name="logger">The <see cref="ILogger"/> instance to use.</param>
+        /// <param name="encryptData">Indicates whether the data should be encrypted using <see cref="IDataProtector"/> while persisted.</param>
+        /// <param name="nameOrConnectionString">The custom connection string or name of the connection string application setting. Used to initialize <see cref="WebHookStoreContext"/>.</param>
+        /// <param name="schemaName">The custom name of database schema. Used to initialize <see cref="WebHookStoreContext"/>.</param>
+        /// <param name="tableName">The custom name of database table. Used to initialize <see cref="WebHookStoreContext"/>.</param>
+        /// <returns></returns>
```
dougbu: Fill in the `<returns>` element. This is one of the reasons the AppVeyor build failed.

src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/WebHooks/SqlWebHookStore.cs
```diff
@@ -66,21 +111,54 @@ public static IWebHookStore CreateStore(ILogger logger)
...
             else
             {
-                store = new SqlWebHookStore(settings, logger);
+                store = new SqlWebHookStore(settings, logger, nameOrConnectionString, schemaName, tableName);
             }
             return store;
         }
 
+        /// <inheritdoc />
+        protected override WebHookStoreContext GetContext()
```
dougbu: Move this after `CheckSqlStorageConnectionString(...)`. This ordering is the other reason Code Analysis failed in the AppVeyor build.

src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/WebHooks/SqlWebHookStore.cs
```diff
@@ -33,6 +37,24 @@ public SqlWebHookStore(SettingsDictionary settings, ILogger logger)
...
+        /// </summary>
+        public SqlWebHookStore(SettingsDictionary settings, ILogger logger, string nameOrConnectionString, string schemaName, string tableName)
+            : base(logger)
+        {
+            if (settings == null)
+            {
+                throw new ArgumentNullException(nameof(settings));
+            }
+
+            _nameOrConnectionString = nameOrConnectionString;
```
dougbu: Sorry, just realized this and the other new constructor overload should call `CheckSqlStorageConnectionString(settings)` when `nameOrConnectionString==null`. Now using the longer overloads all the time in `HttpConfigurationExtensions`.

src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/WebHooks/WebHookStoreContext.cs
```diff
@@ -13,15 +16,58 @@ namespace Microsoft.AspNet.WebHooks
...
+        private readonly string _tableName;
+        private readonly string _schemaName = "WebHooks";
 
         /// <summary>
         /// Initializes a new instance of the <see cref="WebHookStoreContext"/> class.
         /// </summary>
-        public WebHookStoreContext() : base(ConnectionStringNameParameter)
+        public WebHookStoreContext() : base("name=" + ConnectionStringName)
         {
         }
```
dougbu: Up to you: I suggest setting `NameOrConnectionString` in this constructor overload too.

test/Microsoft.AspNet.WebHooks.Custom.SqlStorage.Test/Extensions/HttpConfigurationExtensionsTests.cs
```diff
@@ -22,5 +22,40 @@ public void Initialize_SetsStore()
...
+        {
+            // Arrange
+            HttpConfiguration config = new HttpConfiguration();
+
+            // Act
+            config.InitializeCustomWebHooksSqlStorage(true, nameOrConnectionString);
+            IWebHookStore actual = CustomServices.GetStore();
+
+            // Assert
+            Assert.IsType<SqlWebHookStore>(actual);
```
dougbu: Add a `string` parameter for the `expectedNameOrConnectionString` and `Assert.Equal(expectedNameOrConnectionString, actual.NameOrConnectionString)`. You may need to use the `IsType<T>(...)` return value e.g.
``` c#
var actualStore = Assert.IsType<SqlWebHookStore>(actual);
Assert.Equal(expectedNameOrConnectionString, actualStore.NameOrConnectionString);
```

src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/WebHooks/WebHookStoreContext.cs
```diff
@@ -13,15 +16,58 @@ namespace Microsoft.AspNet.WebHooks
...
+        /// </summary>
+        /// <param name="nameOrConnectionString">Either the database name or a connection string.</param>
+        /// <param name="schemaName">Either the schema name.</param>
+        /// <param name="tableName">Either the table name.</param>
+        public WebHookStoreContext(string nameOrConnectionString, string schemaName, string tableName)
+            : this(nameOrConnectionString)
+        {
+            if (string.IsNullOrEmpty(schemaName))
+            {
+                string msg = string.Format(CultureInfo.CurrentCulture, SqlStorageResources.SqlStore_EmptyString, nameof(schemaName));
```
dougbu: Use `var` as much as possible. This PR is generally pretty good about this but please scan from here to the end.

test/Microsoft.AspNet.WebHooks.Custom.SqlStorage.Test/Extensions/HttpConfigurationExtensionsTests.cs
Marusyk: But `SqlWebHookStore` doesn't have public property `NameOrConnectionString`, there is:
```C#
private readonly string _nameOrConnectionString;
```

Marusyk:
Thank you very much for this review. It was great. I got a great experience. Hope this feature will be helpful for other consumers. Thanks for your time and sorry for so many fixes

Marusyk:
Oh now, what it that?

dougbu:
@Marusyk don't worry about it. FYI there were two instances of `_nameOrConnectionString` versus `nameOrConnectionString`.

Marusyk:
I have to fix it

dougbu:
6c173a343f

Issue #178, opened 2017-11-10
Get rid of TODO and `???` comments


dougbu:
Have filed a few issues to replace TODO items and may file a few more. But, the TODO items need to be cleaned up soon.

The `???` comments are basically questions to reviewers. That got out of hand when #153 grew to almost a hundred commits and almost 250 files. Need to make calls on whatever design decisions aren't yet resolved.

Related points to clean up here:
- Get rid of the `IConfiguration.IsTrue(...)` extension method. Choices include
  1. just use an existing switch e.g. using only `IHostingEnvironment.IsDevelopement()` to disable HTTPS checks.
  2. remove a `bool` knob entirely e.g. switch Stripe to verify request signatures and remove both exiting options, or
  3. map a knob into an `IOptions<T>` class e.g. use `IConfigureOptions<T>` to enable passing Stripe test events through to user actions (allow a setting in configuration as well as user `Startup` code).

  The above choices aren't set in stone for any of the three existing `bool` configuration settings. Might for example choose to get rid of HTTP support all up.
- Redo the `AddSingletonFilter(...)` (now `AddWebHookSingletonFilter(...)` extension methods to simplify the implementation and remove additional extension methods. One option here is to use an `IConfigureOptions<MvcOptions, TFilter>` implementation instead of the bulk of the current method.

dougbu:
Note: Only a few of the TODO and `???` items need design 😈

dougbu:
A few bits addressed as part of other issues. In addition 714a5cd3ec added `IMvcBuilder` extensions and addressed a TODO.

dougbu:
Removing **needs design** label since we've answered the open questions.

dougbu:
Progress&hellip;
- 1585020b74
- e20b745f8a
- fc05979ecc
- 93ea86c77f

dougbu:
e5b1a66654

dougbu:
42117af077

dougbu:
428d143c30

dougbu:
04d1e1ab0f

PR #199, opened 2017-12-04
Infrastructure: Pack Core projects, move to 2.0.0-preview1, and support local incremental versions

dougbu:
- incremental version fix is part of #178
- copy some settings into WebHooks.msbuild for consistent `$(AssemblyRevision)`
- add target to generate unique `$(BuildNumber)` for command-line builds
- use version.props in Packages.proj

Also fix CodeAnalysis build; was duplicating Debug build

nits:
- remove ??? comments and associated commented-out code (if we find problems, we'll fix them)
- rename version.xml to version.props


WebHooks.msbuild
```diff
@@ -16,15 +16,28 @@
...
+    <_TwoDigitYear>$([MSBuild]::Subtract($([System.DateTime]::UtcNow.Year), 2000))</_TwoDigitYear>
+    <_ThreeDigitDayOfYear>$([System.DateTime]::UtcNow.DayOfYear.ToString().PadLeft(3, '0'))</_ThreeDigitDayOfYear>
+    <AssemblyRevision>$(_TwoDigitYear)$(_ThreeDigitDayOfYear)</AssemblyRevision>
   </PropertyGroup>
 
   <ItemGroup>
     <SolutionsToBuild Include="WebHooks.sln">
       <BuildInParallel>$(BuildInParallel)</BuildInParallel>
     </SolutionsToBuild>
-    <NugetsToBuild Include="**\*.nuspec"/>
```
dougbu: Was not passed around at all.

WebHooks.msbuild
```diff
@@ -16,15 +16,28 @@
...
   </ItemGroup>
 
+  <Target Name="GetBuildNumber">
+    <!-- Create second-based build number for local builds. 635556672000000000 is Jan 1, 2015. -->
+    <PropertyGroup>
+      <_SecondBasedTimeStamp>$([System.DateTime]::UtcNow.Subtract($([System.DateTime]::FromBinary(635556672000000000))).TotalSeconds.ToString("F0"))</_SecondBasedTimeStamp>
+      <_SecondBasedTimeStamp>t$([System.Int64]::Parse($(_SecondBasedTimeStamp)).ToString("x9"))</_SecondBasedTimeStamp>
+    </PropertyGroup>
+    <Message Text="/p:BuildNumber=$(_SecondBasedTimeStamp)" Importance="High" />
```
dougbu: If you want `-t000` packages, just run `.\build.cmd`. But if you want packages that can be installed locally, run `.\build.cmd GetBuildNumber` and copy-n-paste its output after `.\build.cmd BuildPackages`.

src/Common/CommonAssemblyInfo.cs
```diff
@@ -26,8 +26,8 @@
 
 #if ASPNETWEBHOOKS
 #if !BUILD_GENERATED_VERSION
-[assembly: AssemblyVersion("1.1.0.0")]
-[assembly: AssemblyFileVersion("1.1.0.0")]
+[assembly: AssemblyVersion("2.0.0.0")]
+[assembly: AssemblyFileVersion("2.0.0.0")]
```
dougbu: This file is used in the old-world projects. For now, they won't have the same assembly attributes e.g. no `[assembly: AssemblyInformationalVersion("2.0.0-preview1-t000")]` or `[assembly: AssemblyMetadata("BuildNumber", "t000")]`.

version.props
```diff
@@ -0,0 +1,10 @@
+<Project>
+  <PropertyGroup>
+    <VersionPrefix>2.0.0</VersionPrefix>
```
dougbu: Yup, I copied this from MVC then changed the `$(VersionPrefix)` 😈

WebHooks.msbuild
```diff
@@ -62,25 +75,26 @@
 
   <Target Name="RestorePackages" DependsOnTargets="DownloadNuGet">
     <Message Text="Restoring NuGet packages..." Importance="High" />
-    <Exec Command='"$(NuGetExe)" restore "%(SolutionsToBuild.Identity)" -PackagesDirectory packages -NonInteractive -Verbosity quiet -ConfigFile "$(MsBuildThisFileDirectory)\.nuget\NuGet.Config"' />
+    <Exec Command='SET BuildNumber=$(BuildNumber)
```
pranavkm: Could these use the restore target? I think it also works on old-type csproj files now (if you have the relevant SDKs installed).

dougbu: Was thinking about that. I'll test it out&hellip;

If it works, do you want to see the new version before I merge into `dev`?

dougbu: (Not sure why this comment got hidden as outdated. I only moved an XML comment from a couple of lines down.)

Experiment failed. The `<restore/>` target supports only `<PackageReference />` and `<ProjectReference />` in the project file. https://docs.microsoft.com/en-us/nuget/schema/msbuild-targets#restore-target isn't very explicit about this but other sources are.

dougbu:
FYI :up::date: just moved a comment

dougbu:
1d8c463e95

Issue #163, opened 2017-10-04
Support Stripe request signatures


dougbu:
A callback is no longer what https://stripe.com/docs/webhooks, a recommends. That page doesn't mention callbacks at all (though the https://api.stripe.com/v1/events/&hellip; API remains).

Should probably use `Stripe-Signature` as described in [this section](https://stripe.com/docs/webhooks#signatures) or (more specifically) [this section](https://stripe.com/docs/webhooks#verify-manually). https://stripe.com/docs/webhooks#replay-attacks is also relevant.

dougbu:
Likely to do this enhancement in the Core world only.

dougbu:
Do this for both receivers. Remove support for existing security verifications.

dougbu:
Turns out the Core Stripe receiver has been broken since #187 went in. Ooops&hellip;

dougbu:
93aa3b3f16

PR #204, opened 2017-12-16
Support Stripe request signatures

dougbu:
- #163
- configuration values are now per-WebHook "signing secrets", not "secret keys", in Stripe terms
  - this is a breaking change; rebuilt applications will need updated configurations
- remove notion of "direct" Stripe requests
- add first Core tests!

supporting changes:
- rewind after reading request `Body` in filters
- fix bug in `TrimmingTokenizer`'s `maxCount` handling
- copy some test infrastructure from aspnet/AspNetWebStack
- redo `StripeTestEventResponseFilter` / `StripeVerifyNotificationIdFilter` split
  - `StripeTestEventResponseFilter` now does all test event handling, including logging
- add `prefix` support to `WebHookVerifyBodyContentFilter.GetRequestBodyHash_SHA256(...)`

nits:
- accept VS suggestions for changed files
- simplify `TrimmingTokenizer` contructors


WebHooks.msbuild
```diff
@@ -115,18 +115,31 @@
...
 
-    <MSBuild Projects="@(XunitProject)" BuildInParallel="$(TestInParallel)" Targets="Xunit" />
+    <MSBuild Projects="@(_XunitProject)" BuildInParallel="$(TestInParallel)" Targets="Xunit" />
+
+    <!-- Failures in these projects will fail build. But, results will not be included in test run summary. -->
+    <MSBuild
+      BuildInParallel="$(TestInParallel)"
+      Properties="Configuration=$(Configuration);Platform=$(_Platform)"
+      Projects="@(_DirectTestProjects)"
+      Targets="Test" />
```
rynowak: I'm kinda trusting you with the MSBuild changes. I don't know exactly what this solves, but it all seems reasonable.

src/Microsoft.AspNet.WebHooks.Receivers.Stripe/WebHooks/StripeWebHookReceiver.cs
```diff
@@ -2,81 +2,53 @@
...
-    public class StripeWebHookReceiver : WebHookReceiver, IDisposable
+    public class StripeWebHookReceiver : WebHookReceiver
     {
         // Application setting to enable test mode
         internal const string PassThroughTestEvents = "MS_WebHookStripePassThroughTestEvents";
 
-        // Do not verify the WebHook using a follow-up HTTP GET request but instead just use the WebHook as-is.
-        internal const string DirectWebHook = "MS_WebHookStripeDirect";
-
         internal const string RecName = "stripe";
```
rynowak: This probably deserves a better name

src/Microsoft.AspNet.WebHooks.Receivers.Stripe/WebHooks/StripeWebHookReceiver.cs
```diff
@@ -93,40 +65,33 @@ public override string Name
...
+                var data = await ReadAsJsonAsync(request);
 
                 // Get notification ID
-                string notificationId = data.Value<string>("id");
+                var notificationId = data.Value<string>("id");
+                var resolver = context.Configuration.DependencyResolver;
                 if (string.IsNullOrEmpty(notificationId))
                 {
-                    string msg = string.Format(CultureInfo.CurrentCulture, StripeReceiverResources.Receiver_BadBody, "id");
+                    var msg = string.Format(CultureInfo.CurrentCulture, StripeReceiverResources.Receiver_BadBody, "id");
```
rynowak: message

WebHooks.msbuild
dougbu: FYI this runs the `test` target in Core projects. That target has the smarts to work around xUnit MSBuild runner not supporting Core projects when run from full framework `msbuild`.

src/Microsoft.AspNet.WebHooks.Receivers.Stripe/WebHooks/StripeWebHookReceiver.cs
dougbu: I agree but&hellip; This is the pattern used in all full framework i.e. non-Core receivers.

src/Microsoft.AspNet.WebHooks.Receivers.Stripe/WebHooks/StripeWebHookReceiver.cs
dougbu: Old style receiver but sure 🎉on.

src/Microsoft.AspNet.WebHooks.Receivers.Stripe/WebHooks/StripeWebHookReceiver.cs
```diff
@@ -148,75 +113,116 @@ public override string Name
...
+                        CultureInfo.CurrentCulture,
+                        StripeReceiverResources.Receiver_InvalidHeaderFormat,
+                        SignatureHeaderName);
+                    request.GetConfiguration().DependencyResolver.GetLogger().Error(message);
+
+                    var invalidHeader = request.CreateErrorResponse(HttpStatusCode.BadRequest, message);
+                    throw new HttpResponseException(invalidHeader);
+                }
+
+                if (!string.IsNullOrEmpty(keyValuePair[1]))
```
rynowak: Is there a valid where this can happen? Should this be an assert?

src/Microsoft.AspNet.WebHooks.Receivers.Stripe/WebHooks/StripeWebHookReceiver.cs
```diff
@@ -148,75 +113,116 @@ public override string Name
...
-            byte[] challenge = Encoding.UTF8.GetBytes(secretKey + ":");
-            req.Headers.Authorization = new AuthenticationHeaderValue("Basic", EncodingUtilities.ToBase64(challenge, uriSafe: false));
-
-            using (HttpResponseMessage rsp = await _httpClient.SendAsync(req))
+            // 4. Verify that the actual hash matches one of the expected hashes.
+            foreach (var expectedSignature in expectedSignatures)
             {
-                if (!rsp.IsSuccessStatusCode)
+                // While this looks repetitious compared to hex-encoding actualHash (once), a single v1 entry in the
+                // header is the normal case. Expect multiple signatures only when rolling secret keys.
```
rynowak: Quality comment 👍

src/Microsoft.AspNetCore.WebHooks.Receivers.Stripe/StripeWebHookAttribute.cs
```diff
@@ -9,33 +9,39 @@ namespace Microsoft.AspNetCore.WebHooks
...
+    /// <see href="https://stripe.com/docs/connect/webhooks"/> for additional details about Stripe Connect WebHook
     /// requests. And, see <see href="https://stripe.com/docs/api/dotnet#events"/> for additional details about Stripe
     /// WebHook request payloads.
     /// </para>
     /// </summary>
+    /// <remarks>
+    /// <para>
+    /// If the application enables CORS in general (see the <c>Microsoft.AspNetCore.Cors</c> package), apply
+    /// <c>DisableCorsAttribute</c> to this action. If the application depends on the
+    /// <c>Microosft.AspNetCore.Mvc.ViewFeatures</c> package, apply <c>IgnoreAntiforgeryTokenAttribute</c> to this
```
rynowak: typo Microosft

src/Microsoft.AspNet.WebHooks.Receivers.Stripe/WebHooks/StripeWebHookReceiver.cs
dougbu: I was mistaken about `SplitAndTrim(...)`. But, I lean toward just removing this check.

dougbu:
93aa3b3f16

PR #206, opened 2017-12-21
Fix problems in Salesforce receiver

dougbu:
- part of #178 and get this receiver working
- fix `SalesforceNotificationsModelBinder` and `SalesforceNotificationsModelBinderProvider`
- remove `SalesforceAcknowledgmentFilter` and update sample to use `ISalesforceResultCreator`
- use correct status code in `SalesforceResultCreator.GetSuccessResultAsync()`

nits:
- improve some doc comments


samples/SalesforceCoreReceiver/Controllers/SalesforceController.cs
```diff
@@ -9,29 +10,31 @@ namespace SalesforceCoreReceiver.Controllers
     public class SalesforceController : ControllerBase
     {
         private readonly ILogger _logger;
+        private readonly ISalesforceResultCreator _resultCreator;
 
-        public SalesforceController(ILoggerFactory loggerFactory)
+        public SalesforceController(ILoggerFactory loggerFactory, ISalesforceResultCreator resultCreator)
```
rynowak: I'm not familiar with the `ISalesforceResultCreator` - what does this do? Are developers expected to customize it?

dougbu: We discussed `ISalesforceResultCreator` a while back. It's the Thingamajig that knows how to create SOAP responses. Source for interface is [here](https://github.com/aspnet/WebHooks/blob/93aa3b3f16d0f10bda9ca320276de68f36869518/src/Microsoft.AspNetCore.WebHooks.Receivers.Salesforce/ISalesforceResultCreator.cs) and implementation is [here](https://github.com/aspnet/WebHooks/blob/93aa3b3f16d0f10bda9ca320276de68f36869518/src/Microsoft.AspNetCore.WebHooks.Receivers.Salesforce/Internal/SalesforceResultCreator.cs).

Details:
1. Users can replace the `ISalesforceResultCreator` with whatever implementation they like. Or, users can ignore it and do something else to create valid SOAP responses. 
2. The `interface` is intended to simplify the Salesforce actions users write. But, [`SalesforceVerifyOrganizationIdFilter`](https://github.com/aspnet/WebHooks/blob/93aa3b3f16d0f10bda9ca320276de68f36869518/src/Microsoft.AspNetCore.WebHooks.Receivers.Salesforce/Filters/SalesforceVerifyOrganizationIdFilter.cs) uses `ISalesforceResultCreator` to handle a few late error cases.
3. Though the implementation in an `Internal` namespace, users _could_ subclass `SalesforceResultCreator` and override a couple of `virtual` methods. I left it in this middle ground state because the exact user requirements aren't clear yet.

So, users have a few options w.r.t. customization and replacement. But, we're not currently promising to keep the implementation as-is.

dougbu:
:up::date: -- rebase only

samples/SalesforceCoreReceiver/Controllers/SalesforceController.cs
rynowak: OK cool 👍

dougbu:
fc05979ecc

PR #205, opened 2017-12-20
Move code around: Refactor and rename filters and metadata interfaces

dougbu:
- part of #178 and #179
- related to #181 because Trello receiver needs secret key confirmation but no challenge
- `WebHookGetResponseFilter` -> `WebHookGetRequestFilter`
  - unconditionally confirm secret key exists
  - stop using `WebHookResultUtilities` in this class
- `WebHookPingResponseFilter` -> `WebHookPingRequestFilter`
- `WebHookVerifyBodyContentFilter` -> `WebHookVerifySignatureFilter` and use it less
  - stop using `EncodingUtilities` or `WebHookResultUtilities` in this class
- `StripeTestEventResponseFilter` -> `StripeTestEventRequestFilter`
- `IWebHookRequestMetadata` -> `IWebHookBodyTypeMetadata`
- `IWebHookRequestMetadataService` -> `IWebHookBodyTypeMetadataSerivce`
- split `IWebHookEventMetadata` into that plus `IWebHookPingRequestMetadata`
- split `IWebHookSecurityMetadata` into `IWebHookGetRequestMetadata` and `IWebHookVerifyCodeMetadata`
  - remove `WebHookGetRequest` class

Also
- remove `EncodingUtilities`
- refactor `WebHookMetadataProvider` while updating metadata validations
- improve comments, logging, and error messages in changed files

PR #208, opened 2017-12-22
Remove `WebHookResultUtilities`, `WebHookExceptionFilter` and `WebHookErrorKeys`

dougbu:
- part of #178
- remove ability to serialize `Exception` details; instead, use `BadRequestObjectResult` directly
- no need for `WebHookExceptionFilter` without `includeErrorDetail` concept

Also
- move `UnixTimeConverter` to Stripe receiver


src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookGetResponseFilter.cs
```diff
@@ -142,7 +141,7 @@ public void OnResourceExecuted(ResourceExecutedContext context)
                     CultureInfo.CurrentCulture,
                     Resources.General_MissingQueryParameter,
                     getMetadata.ChallengeQueryParameterName);
-                var noChallenge = WebHookResultUtilities.CreateErrorResult(message);
+                var noChallenge = new BadRequestObjectResult(message);
```
dougbu: This overlaps #205 because I'd already fixed up the renamed files. Doesn't really matter which goes in first&hellip;

src/Microsoft.AspNetCore.WebHooks.Receivers/WebHookConstants.cs
```diff
@@ -57,6 +57,13 @@ public static class WebHookConstants
             ReceiverConfigurationSectionKey,
             "DisableHttpsCheck");
 
+        /// <summary>
+        /// Gets the model name for the root of body-bound objects. Places errors for <c>data</c> parameters in
+        /// consistent <see cref="Mvc.ModelBinding.ModelStateDictionary"/> entries, separate from entries for (for
+        /// example) parameters bound to route values.
+        /// </summary>
+        public static string ModelStateBodyModelName => "$Body";
```
rynowak: can we make this lowercase?

dougbu: Curious: MSD is case-insensitive. Why worry about the case of this key?

rynowak: because the primary use of this is to serialize it as JSON, and JSON is typically camelCase

dougbu: :ok: then

dougbu:
e20b745f8a

PR #212, opened 2017-12-23
Address some `// ???` comments

dougbu:
- part of #178
- move `TrimmingTokenizer` to shared source
- move most `WebHookReceiverConfigurationExtensions` methods to `WebHookSecurityFilter`
  - move remaining `WebHookReceiverConfigurationExtensions` to shared source
  - rename to `WebHookConfigurationExtensions`
- move most `WebHookHttpRequestUtilities` methods to `WebHookVerifyBodyTypeFilter`
  - stop checking `IsJson()` or `IsXml()` elsewhere
  - move `PrepareRequestBody(...)` to `WebHookHttpRequestUtilities`
  - make `WebHookHttpRequestUtilities` `internal`
- improve some logging
  - e.g. refactor `WebHookEventMapperConstraint` slightly for more specific messages
- `throw` in a few more cases
- remove questions we've discussed and answered


src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
```diff
@@ -125,6 +132,70 @@ public void OnResourceExecuted(ResourceExecutedContext context)
...
+        /// <see langword="true"/> if the specified request contains JSON content; otherwise, <see langword="false"/>.
+        /// </returns>
+        protected static bool IsJson(HttpRequest request)
+        {
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+
+            var contentType = request.GetTypedHeaders()?.ContentType;
```
rynowak: When would this ever return null?

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
```diff
@@ -125,6 +132,70 @@ public void OnResourceExecuted(ResourceExecutedContext context)
...
+            {
+                return false;
+            }
+
+            if (contentType.IsSubsetOf(ApplicationXmlMediaType) || contentType.IsSubsetOf(TextXmlMediaType))
+            {
+                return true;
+            }
+
+            // MVC's XML input formatters do not support text/*+xml by default.
```
rynowak: FYI, this is because the [RFC](https://tools.ietf.org/html/rfc6839#section-3.1) doesn't mention text/json and text/xml. The suffix content type feature specifically applies to application/json and application/xml (among others) specifically

src/shared/Extensions/WebHookConfigurationExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+                throw new ArgumentNullException(nameof(configuration));
+            }
+
+            var value = configuration[key];
+            if (string.IsNullOrEmpty(value))
+            {
+                return false;
+            }
+
+            return bool.TryParse(value.Trim(), out var isSet) ? isSet : false;
```
rynowak: Do we really need this method at all? Seeing things like this in code makes my brain hurt.

It looks to my like TryParse already does what you want. https://msdn.microsoft.com/en-us/library/system.boolean.tryparse(v=vs.110).aspx

I also don't think we should be trimming whitespace around configuration values arbitrarily. Garbage in, garbage out.

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
dougbu: FYI, I know 😈 I added this comment because the ASP.NET WebHooks code checks for the (non-existent) `text/*+json` and `text/*+xml` cases. Don't want some later Doug to make them consistent. Of, if we do make them consistent, change the older code.

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
dougbu: I probably checked for `GetTypedHeaders()` returning `null` based on the older Web API code this morphed from. I'll remove the question mark.

Or, were you asking about the next line's `null` check of `contentType`?

src/shared/Extensions/WebHookConfigurationExtensions.cs
dougbu: Some BCL `Try...(...)` methods `throw` on `null` or empty inputs. I mistakenly thought `bool.TryParse(...)` was one of those. So, yeah, I'll remove this class.

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
```diff
@@ -125,6 +132,74 @@ public void OnResourceExecuted(ResourceExecutedContext context)
...
+            }
+
+            if (contentType.IsSubsetOf(ApplicationJsonMediaType) || contentType.IsSubsetOf(TextJsonMediaType))
+            {
+                return true;
+            }
+
+            // MVC's JsonInputFormatter does not support text/*+json by default. RFC 3023 and 6839 allow */*+json but
+            // https://www.iana.org/assignments/media-types/media-types.xhtml shows all +json registrations except
+            // model/gltf+json match application/*+json.
```
dougbu: > FYI, this is because the [RFC](https://tools.ietf.org/html/rfc6839#section-3.1) doesn't mention text/json and text/xml. The suffix content type feature specifically applies to application/json and application/xml (among others) specifically

I've updated this comment.

More generally, I'm not sure limiting ourselves to `application/*+json` and `application/*+xml` is correct according to [RFC 6839](https://tools.ietf.org/html/rfc6839). That and [RFC 3023](https://tools.ietf.org/html/rfc3023) talk more in terms of `*/*+json` and `*/*+xml` though I agree media types registered with suffixes (so far) almost all match `application/*+json` or `application/*+xml`. (The current exceptions match `message/*+xml`, `model/*+xml` and `model/gltf+json`.)

Thoughts about filing an aspnet/Http or aspnet/Mvc bug to further generalize of suffix handling?

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
```diff
@@ -125,6 +132,74 @@ public void OnResourceExecuted(ResourceExecutedContext context)
...
+        /// </returns>
+        protected static bool IsJson(HttpRequest request)
+        {
+            if (request == null)
+            {
+                throw new ArgumentNullException(nameof(request));
+            }
+
+            var contentType = request.GetTypedHeaders().ContentType;
+            if (contentType == null)
```
dougbu: > @rynowak: When would this ever return null?
> @dougbu: &hellip;
> Or, were you asking about the next line's null check of contentType?

@rynowak just double-checking: Is this now fine w/ you?

src/Microsoft.AspNetCore.WebHooks.Receivers.Stripe/Filters/StripeTestEventRequestFilter.cs
```diff
@@ -71,7 +71,8 @@ public void OnResourceExecuting(ResourceExecutingContext context)
             if (IsTestEvent(notificationId))
             {
                 // Log about and optionally short-circuit this test event.
-                if (_configuration.IsTrue(StripeConstants.PassThroughTestEventsConfigurationKey))
+                var passThroughString = _configuration[StripeConstants.PassThroughTestEventsConfigurationKey];
+                if (bool.TryParse(passThroughString, out var passThrough) && passThrough)
```
dougbu: @rynowak I'm assuming you were objecting to the helper method. But, is this the syntax that makes your head hurt i.e. do you dislike everything but
``` c#
var passThroughString = _configuration[StripeConstants.PassThroughTestEventsConfigurationKey];
bool.TryParse(passThroughString, out var passThrough);
if (passThrough)
{
    // ...
}
```

rynowak: I am objecting the helper method.

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
rynowak: 👍

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
rynowak: If you want to create a paper trail, feel free to create one. I think we'd decide to close it.

We register default content types for the common use cases, not all possible cases. Application developers can do whatever they want 👍

dougbu:
>We made steak nachos for dinner and it has given one of our dogs the NACHO CRAZIES

Good to know but needs emojis: Dinner 🍽 for dogs 🐶 often dog 🐕 food 🥘 -- sometimes expensive 💰 dog 🐩 food 🥘

dougbu:
e5b1a66654

PR #213, opened 2017-12-31
Add `TrimmingTokenizer` and `UnixTimeConverter` tests

dougbu:
- start from ASP.NET WebHooks data sets; see `StringExtensionsTests` and `UnixTimeConverterTests`
- also add missing argument checks in `TrimmingTokenizer`

### DIFF
--- a/src/shared/Utilities/TrimmingTokenizer.cs
+++ b/src/shared/Utilities/TrimmingTokenizer.cs
@@ -14,17 +14,6 @@ namespace Microsoft.AspNetCore.WebHooks.Utilities
     /// </summary>
     internal struct TrimmingTokenizer : IEnumerable<StringSegment>
     {
-        private static readonly char[] WhitespaceCharacters = new[]
-        {
-            // Cribbed from a char.IsWhitespace() optimization in
-            // https://www.codeproject.com/Articles/1014073/Fastest-method-to-remove-all-whitespace-from-Strin
-            '\u0009', '\u000A', '\u000B', '\u000C', '\u000D',
-            '\u0020', '\u0085', '\u00A0', '\u1680', '\u2000',
-            '\u2001', '\u2002', '\u2003', '\u2004', '\u2005',
-            '\u2006', '\u2007', '\u2008', '\u2009', '\u200A',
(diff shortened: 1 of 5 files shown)


src/shared/Utilities/TrimmingTokenizer.cs
```diff
@@ -14,6 +14,17 @@ namespace Microsoft.AspNetCore.WebHooks.Utilities
...
+        private static readonly char[] WhitespaceCharacters = new[]
+        {
+            // Cribbed from a char.IsWhitespace() optimization in
+            // https://www.codeproject.com/Articles/1014073/Fastest-method-to-remove-all-whitespace-from-Strin
+            '\u0009', '\u000A', '\u000B', '\u000C', '\u000D',
+            '\u0020', '\u0085', '\u00A0', '\u1680', '\u2000',
+            '\u2001', '\u2002', '\u2003', '\u2004', '\u2005',
+            '\u2006', '\u2007', '\u2008', '\u2009', '\u200A',
+            '\u2028', '\u2029', '\u202F', '\u205F', '\u3000',
+        };
```
rynowak: Why do we need this? I don't understand why we need a general purpose tokenizer in webhooks. Most of these are for textual markup and would never be used in the cases we care about?

dougbu: I'm not sure and haven't explored all uses of the `string.SplitAndTrim(...)` extension method in the ASP.NET WebHooks code. However, the tests of that method definitely tested the `null` or empty `separators` case.

More importantly, I needed to choose reasonable behaviour for this case. This default and alignment with `string.Split(...)`, `string.Trim(...)`, &hellip; was much simpler than making this an error condition and `throw`ing, especially when I considered a need to localize messages for this shared code.

dougbu:
:up::date:

dougbu:
This PR is now primarily about adding tests for `TrimmingTokenizer` and `UnixTimeConverter`. Changes to `TrimmingTokenizer` are limited to argument checks.

Will update title and description tomorrow.

dougbu:
9df8808679

Issue #185, opened 2017-11-11
Add `WebHookEventMapperFilter`


dougbu:
Follow-on work item for #5.

New ASP.NET Core receivers lack support for requiring event names found in the request body. Those receivers that have "body-sourced" event names should also support model binding those values.

Self-assigning because I'm already working on this. It's a feature gap I've mentioned a few times&hellip;

dougbu:
Underlying pieces in 428d143c30

dougbu:
b4e374bf43

PradeepLoganathan:
Is there a tentative timeline when webhooks will be available on .net core please ?

PR #215, opened 2018-01-01
Add `IWebHookRequestReader` and `WebHookRequestReader`

dougbu:
- part of #178 (offline conversations) and support for #185
- change `RouteData.SetWebHookEventNames(...)` to take names as `StringValues`
- do not instantiate `SalesforceNotifications` in Salesforce filter
  - add `ObjectPathUtilities.GetStringValues(...)` methods and use in that filter
- do not use model binding to get `IFormCollection` in Slack filter


src/Microsoft.AspNetCore.WebHooks.Receivers/Extensions/WebHookRequestReaderExtensions.cs
```diff
@@ -0,0 +1,298 @@
...
+        /// Read the HTTP request entity body (formatted as HTML form URL-encoded data) as an
+        /// <see cref="IFormCollection"/> instance.
+        /// </summary>
+        /// <param name="requestReader">The <see cref="IWebHookRequestReader"/> this method extends.</param>
+        /// <param name="actionContext">The <see cref="ActionContext"/> for the current request and action.</param>
+        /// <returns>
+        /// A <see cref="Task"/> that on completion provides an <see cref="IFormCollection"/> instance containing data
+        /// from the HTTP request entity body.
+        /// </returns>
+        public static async Task<IFormCollection> ReadAsFormDataAsync(
```
rynowak: What do we we need this for? As far as I know the the form data code already does all of these things

src/Microsoft.AspNetCore.WebHooks.Receivers/IWebHookRequestReader.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        /// <summary>
+        /// Get an indication whether the <paramref name="request"/> is suitable to be read by
+        /// <see cref="ReadBodyAsync{TModel}"/> or a similar request entity body reader.
+        /// </summary>
+        /// <param name="request">The <see cref="HttpRequest"/> to check.</param>
+        /// <returns>
+        /// <see langword="true"/> if the <paramref name="request"/> is suitable to be read by
+        /// <see cref="ReadBodyAsync{TModel}"/>; <see langword="false"/> otherwise.
+        /// </returns>
```
rynowak: What does this mean though? What is `suitable` and `valid`

src/Microsoft.AspNetCore.WebHooks.Receivers/IWebHookRequestReader.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <param name="valueProviderFactories">
+        /// The collection of configured <see cref="IValueProviderFactory"/>s.
+        /// </param>
+        /// <returns>
+        /// A <see cref="Task"/> that on completion provides a <typeparamref name="TModel"/> instance containing the
+        /// HTTP request entity body.
+        /// </returns>
+        Task<TModel> ReadBodyAsync<TModel>(
+            ActionContext actionContext,
+            IList<IValueProviderFactory> valueProviderFactories);
```
rynowak: Can you give an example of how the value provider factories are used?

src/Microsoft.AspNetCore.WebHooks.Receivers/Extensions/WebHookRequestReaderExtensions.cs
dougbu: The form data code neither ensures a re-settable stream nor resets the `Body` stream's position.

src/Microsoft.AspNetCore.WebHooks.Receivers/IWebHookRequestReader.cs
dougbu: This isn't incredibly real, more an MVC infrastructure requirement, because `BodyModelBinder` doesn't use it. Need some non-`null` `IValueProvider` to call `DefaultModelBindingContext.CreateBindingContext(...)` (the constructors don't ensure the properties have consistent values) and I chose to use `CompositeValueProvider.CreateAsync(...)` because the `IList<IValueProviderFactory>` is available everywhere I expect this interface to be used.

One option if I remove the list from `IWebHookRequestReader` would be for `WebHookRequestReader` to simply construct an empty `CompositeValueProvider`. That type's default constructor is `public`. Make sense?

FYI I was using the `ModelBinderFactory` at one point. That meant the reader was not guaranteed to use a `BodyModelBinder`. So, this is in part leftover from then.

src/Microsoft.AspNetCore.WebHooks.Receivers/IWebHookRequestReader.cs
dougbu: Short answer is it means `ReadBodyAsync(...)` won't always give up and return `null`. I'll update the comments&hellip;

src/Microsoft.AspNetCore.WebHooks.Receivers/Extensions/WebHookRequestReaderExtensions.cs
dougbu: More specifically, `ReadFormAsync(...)` doesn't always enable rewinding the stream (it's controlled by `FormOptions.BufferBody`), only drains the stream in one case, and rewinds only when the stream allows that. WebHooks needs all that.

src/Microsoft.AspNetCore.WebHooks.Receivers/IWebHookRequestReader.cs
dougbu: No need to update the comments. I removed the method from the `interface`.

dougbu:
:up::date:

Suggest having a look at the 2nd commit description.

src/Microsoft.AspNetCore.WebHooks.Receivers/IWebHookRequestReader.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Microsoft.AspNetCore.WebHooks
+{
+    /// <summary>
+    /// Provides an abstraction for a service that parses the request body. For use in filters that execute prior to
+    /// regular model binding or in actions that read the request body after regular model binding.
+    /// </summary>
+    public interface IWebHookRequestReader
```
rynowak: Consider making this an abstract class. It seems like it might grow. I'm pretty careful now -- if an interface has more than one use case, an abstract class is usually going to work better in the end.

dougbu: @rynowak and I discussed this point offline. Agreed `IWebHookRequestReader` is fine as-is.

dougbu:
428d143c30

PR #218, opened 2018-01-09
Bind event names from request body

dougbu:
- #185 and #178 (addresses a big TODO)
- add `WebHookEventMapperFilter`
- add `IWebHookEventFromBodyMetadata` and implement it in some metadata classes

### DIFF
--- a/samples/PusherCoreReceiver/Controllers/PusherController.cs
+++ b/samples/PusherCoreReceiver/Controllers/PusherController.cs
@@ -16,7 +16,7 @@ namespace PusherCoreReceiver.Controllers
         }
 
         [PusherWebHook(Id = "It")]
-        public IActionResult PusherForIt(PusherNotifications data)
+        public IActionResult PusherForIt(string[] events, PusherNotifications data)
         {
             if (!ModelState.IsValid)
             {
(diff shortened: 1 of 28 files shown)


samples/PusherCoreReceiver/Controllers/PusherController.cs
```diff
@@ -87,17 +96,27 @@ public IActionResult Pusher(string id, JObject data)
             // Get the notification creation timestamp.
             var createdAtUnix = data.Value<long>(PusherConstants.EventRequestCreatedAtPropertyName);
             var createdAt = DateTimeOffset.FromUnixTimeMilliseconds(createdAtUnix);
-            var events = data.Value<JArray>(PusherConstants.EventRequestPropertyContainerName);
+            var eventArray = data.Value<JArray>(PusherConstants.EventRequestPropertyContainerName);
```
rynowak: What's the difference between these and `events`? Should like have a different name that's more descriptive?

dougbu: This is a single `JArray` containing all of the event objects. The `eventEnumerable` and individual `@event` objects and `eventName` strings come from this `JArray`.

`events` is an array bound to the event names found in the request. I chose the parameter name for consistency across the samples. But, I could rename that one to `eventNames` if you like because WebHooks binds all of `@event`, `events`, `eventName`, and `eventNames` plus `action` through `actionNames`. See https://github.com/aspnet/WebHooks/blob/428d143c300af46f83ba4149c86b5d5f8953065f/src/Microsoft.AspNetCore.WebHooks.Receivers/ApplicationModels/WebHookModelBindingProvider.cs#L91-L105

 (`eventName`s shouldn't be found as shown below in a real app because all the names are passed in the `@events` argument. I'm only checking that property in this sample to double-check the argument. That is, model binding finds all of the names of the event objects.)

rynowak: Maybe `eventNames` for the `string[]` and `events` for the  `JArray`?

dougbu:
b4e374bf43

PR #219, opened 2018-01-09
Add missing variable for structured logging in signature filter

twsouthwick:
The ASP.NET Core logging infrastructure is structured logging, so for each brace-escaped name, a variable must be given. This is in contrast to `string.Format` style logging where the braces indicate an index. Without this fix, an IndexOutOfRangeException is thrown and the user just sees a 500 error.


dougbu:
Thanks @twsouthwick. Good catch!

Issue #222, opened 2018-01-11
Split solution in two


dougbu:
- move ASP.NET WebHooks projects to AspNetWebHooks.sln
- remove `<Compile>/<Link>`s from ASP.NET Core projects to ASP.NET projects; copy the files instead

dougbu:
Self-assigned and started work&hellip;

dougbu:
febed48de3

PR #224, opened 2018-01-15
Split the solution in two

dougbu:
- #222
- remove links from ASP.NET Core projects to ASP.NET projects
  - copy the C# files over
- remove conditional directives from files that are no longer shared
- rename .msbuild file and focus infrastructure on building AspNetWebHooks.sln
  - will use KoreBuild for ASP.NET Core WebHooks projects

nit: accept VS suggestions in changed files


src/Microsoft.AspNetCore.WebHooks.Receivers.AzureAlert/AzureAlertCondition.cs
```diff
@@ -0,0 +1,73 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.AspNetCore.WebHooks
```
dougbu: Recommend ignoring the new C# files. They are identical to the Microsoft.AspNet.WebHooks versions except for their namespace and (in a couple of cases) `using`s or doc comments. That is, the edits are the opposite to those above except the VS suggestions were identical.

rynowak: OK, I believe you

src/Microsoft.AspNetCore.WebHooks.Receivers.Salesforce/Microsoft.AspNetCore.WebHooks.Receivers.Salesforce.csproj
```diff
@@ -6,18 +6,7 @@
...
-    <EmbeddedResource Include="..\Microsoft.AspNet.WebHooks.Receivers.Salesforce\Messages\NotificationResponse.xml">
-      <Link>Messages\NotificationResponse.xml</Link>
-    </EmbeddedResource>
-    <Compile Include="..\Microsoft.AspNet.WebHooks.Receivers.Salesforce\WebHooks\SalesforceNamespaces.cs">
-      <Link>SalesforceNamespaces.cs</Link>
-    </Compile>
-    <Compile Include="..\Microsoft.AspNet.WebHooks.Receivers.Salesforce\WebHooks\SalesforceNotifications.cs">
-      <Link>SalesforceNotifications.cs</Link>
-    </Compile>
+    <EmbeddedResource Include="Messages\*.xml" />
```
rynowak: 👍

Issue #181, opened 2017-11-11
Flesh out Trello and WordPress ASP.NET Core receivers


dougbu:
Follow-on work item for #5.

dougbu:
One "half": 0fd2961cb3

dougbu:
- d00d67fa13, support short-circuiting HEAD requests
- 547c5202b0, improve `WebHookVerifySignatureFilter` API
  - includes new methods for Trello receiver
- 9b6c58751e, add Trello receiver

PR #226, opened 2018-01-18
Add WordPress receiver

dougbu:
- second part of #181

nit: correct a `[GitHubWebHook]` comment


src/Microsoft.AspNetCore.WebHooks.Receivers.WordPress/Internal/WordPressServiceCollectionSetup.cs
```diff
@@ -0,0 +1,30 @@
...
+using Microsoft.AspNetCore.WebHooks.Metadata;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+
+namespace Microsoft.AspNetCore.WebHooks.Internal
+{
+    /// <summary>
+    /// Methods to add services for the WordPress receiver.
+    /// </summary>
+    public static class WordPressServiceCollectionSetup
```
rynowak: Suggest making this proper internal.

dougbu: Filed #233 to do this in one fell swoop.

Issue #195, opened 2017-11-29
Remove double quotes around URLs in NUSPEC metadata


Eilon:
E.g. here: https://github.com/aspnet/WebHooks/blob/cfb4fe63ae789c602b63bc1f8f6b9a018f044a59/src/Microsoft.AspNet.WebHooks.Receivers.Paypal/Microsoft.AspNet.WebHooks.Receivers.Paypal.nuspec#L13-L15

This causes a problem on the NuGet gallery, which you can see if you click on the links in the description here: https://www.nuget.org/packages/Microsoft.AspNet.WebHooks.Receivers.Paypal/1.2.1

So, we should remove the quotes from around URLs in the NUSPECs of any of the packages here (they aren't really needed).

dougbu:
@Eilon milestone?

Eilon:
Need to check if the trailing `.` that would be at the end of the URL would also cause a problem with the NuGet Gallery. We might want to play it safe and make sure there's a space after each URL to help guarantee that the gallery will parse the URL correctly.

dougbu:
Moving into Preview1 after a quick chat with @mkArtakMSFT

PR #227, opened 2018-01-18
Update URLs in .nuspec and README files

dougbu:
- remove double quotes around URLs in `<description>` (#195)
  - will send another PR (after next build) if a URL then a period looks odd on MyGet
- use HTTPS for all URLs in these files (none were HTTP-only)
  - do the same thing for C# and a few HTML files where target site supports HTTPS
- replace links to sender documentation where they've gotten stale

### DIFF
--- a/src/Microsoft.AspNet.WebHooks.Common/Microsoft.AspNet.WebHooks.Common.nuspec
+++ b/src/Microsoft.AspNet.WebHooks.Common/Microsoft.AspNet.WebHooks.Common.nuspec
@@ -6,9 +6,9 @@
     <title>Microsoft ASP.NET WebHooks Common Module</title>
     <authors>Microsoft</authors>
     <owners>Microsoft, aspnet</owners>
-    <licenseUrl>http://www.microsoft.com/web/webpi/eula/net_library_eula_ENU.htm</licenseUrl>
-    <projectUrl>http://go.microsoft.com/fwlink/?LinkId=690277</projectUrl>
-    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <licenseUrl>https://www.microsoft.com/web/webpi/eula/net_library_eula_enu.htm</licenseUrl>
+    <projectUrl>https://go.microsoft.com/fwlink/?LinkId=690277</projectUrl>
+    <iconUrl>https://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
     <requireLicenseAcceptance>true</requireLicenseAcceptance>
     <description>
       This package provides common functionality for sending and receiving WebHooks using ASP.NET.
--- a/src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/Microsoft.AspNet.WebHooks.Custom.SqlStorage.nuspec
+++ b/src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/Microsoft.AspNet.WebHooks.Custom.SqlStorage.nuspec
@@ -6,9 +6,9 @@
     <title>Microsoft ASP.NET Custom WebHooks SQL Module</title>
     <authors>Microsoft</authors>
     <owners>Microsoft, aspnet</owners>
-    <licenseUrl>http://www.microsoft.com/web/webpi/eula/net_library_eula_ENU.htm</licenseUrl>
-    <projectUrl>http://www.asp.net/</projectUrl>
-    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <licenseUrl>https://www.microsoft.com/web/webpi/eula/net_library_eula_enu.htm</licenseUrl>
+    <projectUrl>https://www.asp.net/</projectUrl>
+    <iconUrl>https://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
     <requireLicenseAcceptance>true</requireLicenseAcceptance>
     <description>
       This package provides support for persisting your custom WebHooks registrations in a SQL database.
--- a/src/Microsoft.AspNet.WebHooks.Receivers.Azure/Microsoft.AspNet.WebHooks.Receivers.Azure.nuspec
+++ b/src/Microsoft.AspNet.WebHooks.Receivers.Azure/Microsoft.AspNet.WebHooks.Receivers.Azure.nuspec
@@ -6,12 +6,14 @@
     <title>Microsoft ASP.NET WebHooks Receiver for Azure WebHooks.</title>
     <authors>Microsoft</authors>
     <owners>Microsoft, aspnet</owners>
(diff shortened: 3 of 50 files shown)


src/Microsoft.AspNet.WebHooks.Common/Microsoft.AspNet.WebHooks.Common.nuspec
```diff
@@ -6,9 +6,9 @@
     <title>Microsoft ASP.NET WebHooks Common Module</title>
     <authors>Microsoft</authors>
     <owners>Microsoft, aspnet</owners>
-    <licenseUrl>http://www.microsoft.com/web/webpi/eula/net_library_eula_ENU.htm</licenseUrl>
-    <projectUrl>http://go.microsoft.com/fwlink/?LinkId=690277</projectUrl>
-    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <licenseUrl>https://www.microsoft.com/web/webpi/eula/net_library_eula_enu.htm</licenseUrl>
+    <projectUrl>https://go.microsoft.com/fwlink/?LinkId=690277</projectUrl>
```
dougbu: This URL redirects to the [ASP.NET WebHooks overview](https://docs.microsoft.com/en-us/aspnet/webhooks/) page in our documentation. We have a build task that replaces this with http://www.asp.net/ (yes, not HTTPS ☹️ -- separate bug) and our ASP.NET Core WebHooks projects use https://asp.net. Which of these three do we prefer in the ASP.NET WebHooks .nuspec files?

src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/Microsoft.AspNet.WebHooks.Custom.SqlStorage.nuspec
```diff
@@ -6,9 +6,9 @@
     <title>Microsoft ASP.NET Custom WebHooks SQL Module</title>
     <authors>Microsoft</authors>
     <owners>Microsoft, aspnet</owners>
-    <licenseUrl>http://www.microsoft.com/web/webpi/eula/net_library_eula_ENU.htm</licenseUrl>
-    <projectUrl>http://www.asp.net/</projectUrl>
-    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <licenseUrl>https://www.microsoft.com/web/webpi/eula/net_library_eula_enu.htm</licenseUrl>
+    <projectUrl>https://www.asp.net/</projectUrl>
```
dougbu: This is unique. I left it alone 'til we decide on my [question above](https://github.com/aspnet/WebHooks/pull/227/files#r162497646).

src/Microsoft.AspNet.WebHooks.Common/Microsoft.AspNet.WebHooks.Common.nuspec
Eilon: The FWLink sounds fine, so long as it actually stays there. Can we fix the thing that overwrites it?

src/Microsoft.AspNet.WebHooks.Custom.SqlStorage/Microsoft.AspNet.WebHooks.Custom.SqlStorage.nuspec
Eilon: I can't imagine any reason why this would need to be different. Let's make it match.

src/Microsoft.AspNet.WebHooks.Receivers.Azure/Microsoft.AspNet.WebHooks.Receivers.Azure.nuspec
```diff
@@ -6,12 +6,14 @@
...
-    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <licenseUrl>https://www.microsoft.com/web/webpi/eula/net_library_eula_enu.htm</licenseUrl>
+    <projectUrl>https://go.microsoft.com/fwlink/?LinkId=690277</projectUrl>
+    <iconUrl>https://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
     <requireLicenseAcceptance>true</requireLicenseAcceptance>
     <description>
-      This package provides support for receiving WebHooks from Kudu (Azure Web App Deployment) and Azure Alerts. For information about Kudu WebHooks, see "https://github.com/projectkudu/kudu/wiki/Web-hooks" and for Azure Alerts, see "https://azure.microsoft.com/en-us/documentation/articles/insights-webhooks-alerts/".
+      This package provides support for receiving WebHooks from Kudu (Azure Web App Deployment) and Azure Alerts. For
+      information about Kudu WebHooks, see https://github.com/projectkudu/kudu/wiki/Web-hooks and for Azure Alerts, see
+      https://azure.microsoft.com/en-us/documentation/articles/insights-webhooks-alerts/.
```
Eilon: BTW whether this shows up correctly on MyGet is unlikely to be a good indication of whether this looks good on NuGet.org. The code for it is here: https://github.com/NuGet/NuGetGallery/blob/da86aef01bf5bfbdaedac486137d0019b75dcdbe/src/NuGetGallery/Helpers/HtmlExtensions.cs#L54

It would be prudent to just try that code on this input and see what it generates.

dougbu: Thanks very much for the pointer. We're good.

Tried it out (using our new friend the VS C# Interactive window 😺) with the following sample:
``` c#
var text = " https://azure.microsoft.com/en-us/documentation/articles/insights-webhooks-alerts/.";
```
Result was just lovely:
``` c#
" <a href=\"https://azure.microsoft.com/en-us/documentation/articles/insights-webhooks-alerts/\" rel=\"nofollow\">https://azure.microsoft.com/en-us/documentation/articles/insights-webhooks-alerts/</a>."
```

src/Microsoft.AspNet.WebHooks.Common/Microsoft.AspNet.WebHooks.Common.nuspec
dougbu: 👍 I've updated a related issue on "the thing" to include this fix.

dougbu:
:up::date:

src/Microsoft.AspNet.WebHooks.Receivers.Azure/Microsoft.AspNet.WebHooks.Receivers.Azure.nuspec
Eilon: Nice!

Issue #231, opened 2018-01-20
Change `SlackCommand.ParseParameters(...)` to not return a `ValueTuple`


dougbu:
The Microsoft.AspNetCore.WebHooks.Receivers.Slack package currently has a `public` method that returns a `ValueTuple`. All consuming applications (including tests) using this method hit dotnet/standard#567 if they target .NET Framework and run on systems with .NET Framework 4.7.1 installed.

Fix is to define a type containing the return values, to use `KeyValuePair<TKey, TValue>`, or to make one of the return values (likely the `string Error`) an `out` parameter.

The other option is to wait for a dotnet/standard#567 fix. But, that's not coming when we expect to ship version 1.0.0-preview1.

dougbu:
I have a fix for this if we decide to take it in preview1.

dougbu:
Try to get this into Preview1. That said, it's an issue in a single receiver only on full .NET Framework.

dougbu:
e65d214c5b

PR #234, opened 2018-01-24
`SlackCommand.ParseParameters(...)` -> `TryParseParameters(...)`

dougbu:
- #231
- avoid returning a `ValueTuple` and hitting dotnet/standard#567
- reenable tests skipped in #232

### DIFF
--- a/src/Microsoft.AspNetCore.WebHooks.Receivers.Slack/SlackCommand.cs
+++ b/src/Microsoft.AspNetCore.WebHooks.Receivers.Slack/SlackCommand.cs
@@ -74,14 +74,19 @@ namespace Microsoft.AspNetCore.WebHooks
         /// <see cref="KeyValuePair{StringSegment, StringSegment}.Value"/> from what
         /// <see cref="ParseActionWithValue"/> returned.
         /// </param>
+        /// <param name="errorMessage">
+        /// A <see cref="string"/> describing the specific problem when an issue is encountered. <see langword="null"/>
+        /// when the method successfully returns a non-<see langword="null"/> dictionary.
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNetCore.WebHooks.Receivers.Slack/SlackCommand.cs
```diff
@@ -74,14 +74,19 @@ public static class SlackCommand
...
-        /// <c>Error</c> is <see langword="null"/> and <c>Parameters</c> an <see cref="IDictionary{TKey, TValue}"/>
-        /// mapping zero or more parameter names to zero or more values. Otherwise, <c>Parameters</c> is
-        /// <see langword="null"/> and <c>Error</c> a <see cref="string"/> describing the specific problem.
+        /// An <see cref="IDictionary{TKey, TValue}"/> mapping zero or more parameter names to zero or more values.
+        /// <see langword="null"/> when an issue is encountered. <paramref name="errorMessage"/> and the return value
+        /// are never both <see langword="null"/>.
         /// </returns>
-        public static (IDictionary<StringSegment, IList<StringSegment>> Parameters, string Error) ParseParameters(
-            StringSegment text)
+        public static IDictionary<StringSegment, IList<StringSegment>> TryParseParameters(
```
rynowak: consider converting this to a `bool TryParseParameters(text, out paramters, out errorMessage)`

I usually expect a bool when I see `Try`. 

I don't think it's super important

dougbu: Went with your not "super important" leaning, partially because the double-`out` part looks icky and partially to avoid redoing the doc comments yet again.

PR #254, opened 2018-02-28
Allow `[GeneralWebHook]` without a required `BodyType`

dougbu:
- make `WebHookBodyType` a `[Flags]` enum and add `All` value
- require `IWebHookBodyTypeMetadata` for all receivers
  - implement `IWebHookBodyTypeMetadata` in `WebHookMetadata` (`abstract`ly)
- for `[GeneralWebHook]`, have `WebHookVerifyBodyTypeFilter` enforce receiver's requirements
- move now-shared methods into `RequestBodyTypes`

nits:
- log warnings in `WebHookModelBindingProvider`
- add and update comments
  - in particular, add more comments about `IApplicationModelProvider.Order`
- refactor validity checks out of `WebHookMetadataProvider.Apply(...)`
- remove unused parameter from `WebHookRoutingProvider.ChooseTemplate()`
- remove leftover file from solution; add dependencies.props
- VS removed byte order marks


dougbu:
Rather large change that I can't really break apart -- sorry. Core is in `WebHookMetadataProvider`, `WebHookEventMapperFilter`, `WebHookVerifyBodyTypeFilter`, and the metadata interfaces.

@Eilon you might want to check out the doc comments since this PR is somewhat comment-heavy.

src/Microsoft.AspNetCore.WebHooks.Receivers/ApplicationModels/WebHookModelBindingProvider.cs
```diff
@@ -11,18 +11,53 @@
...
+    /// An <see cref="IApplicationModelProvider"/> implementation that adds model binding information
+    /// (<see cref="BindingInfo"/> settings similar to <see cref="IBindingSourceMetadata"/> and
+    /// <see cref="IModelNameProvider"/>) to <see cref="ParameterModel"/>s of WebHook actions.
     /// </summary>
     public class WebHookModelBindingProvider : IApplicationModelProvider
     {
+        private readonly ILogger _logger;
+
+        /// <summary>
+        /// Instantiates a new <see cref="WebHookRoutingProvider"/> with the given
```
Eilon: Wrong cref?

src/Microsoft.AspNetCore.WebHooks.Receivers/ApplicationModels/WebHookModelBindingProvider.cs
```diff
@@ -11,18 +11,53 @@
...
+        /// Add routing information (template, constraints and filters) to <see cref="ActionModel"/>s (in
+        /// <see cref="WebHookRoutingProvider"/>).
+        /// </item>
+        /// <item>
+        /// Add model binding information (<see cref="BindingInfo"/> settings) to <see cref="ParameterModel"/>s (in
+        /// this filter).
+        /// </item>
+        /// </list>
+        /// </summary>
+        public static int Order => WebHookRoutingProvider.Order + 10;
```
Eilon: Just wondering, why 10 now instead of 20? And why not 100 or 1,000?

Eilon: And should this be documented?

Eilon: Is user-written stuff meant to interleave with our providers?

src/Microsoft.AspNetCore.WebHooks.Receivers/Metadata/WebHookBodyType.cs
```diff
@@ -1,30 +1,52 @@
...
+        /// Request may have any supported <c>content-type</c>. The
+        /// <see cref="ApplicationModels.WebHookModelBindingProvider"/> ignores bound <c>data</c> parameters when the
+        /// action or receiver has this <see cref="WebHookBodyType"/>.
         /// </summary>
-        Xml,
+        /// <remarks>
+        /// This value is intended for use as <see cref="GeneralWebHookAttribute"/>'s default
+        /// <see cref="IWebHookBodyTypeMetadata.BodyType"/>.
+        /// </remarks>
+        All = Form | Json | Xml,
```
Eilon: BTW it would be a breaking change to add a new value to this enum because the value of `All` would likely change, and these are all `const`s.

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookEventMapperFilter.cs
```diff
@@ -124,9 +125,36 @@ public class WebHookEventMapperFilter : IAsyncResourceFilter
                 return;
             }
 
-            StringValues eventNames;
+            // Determine the applicable WebhookBodyType i.e. how to read the request body.
+            WebHookBodyType bodyType;
+            var request = context.HttpContext.Request;
             var bodyTypeMetadata = _bodyTypeMetadata.First(metadata => metadata.IsApplicable(receiverName));
```
Eilon: Is there always at least one applicable entry in the list? What if there isn't?

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
```diff
@@ -97,118 +128,153 @@ public void OnResourceExecuting(ResourceExecutingContext context)
...
+            if (_allBodyTypeMetadata != null)
+            {
+                var receiverBodyTypeMetadata = _allBodyTypeMetadata
+                    .FirstOrDefault(metadata => metadata.IsApplicable(receiverName));
+                if (receiverBodyTypeMetadata == null)
+                {
+                    // WebHookMetadataProvider requires an IWebHookBodyTypeMetadataService implementation for all
+                    // receivers. But, this case may occur if user has no receiver-specific actions i.e. uses only the
+                    // GeneralWebHookAttribute.
+                    _logger.LogCritical(
```
Eilon: Doesn't really seem *critical* to me... https://github.com/aspnet/Logging/wiki/Guidelines#critical

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
```diff
@@ -97,118 +128,153 @@ public void OnResourceExecuting(ResourceExecutingContext context)
...
+                {
+                    // Use receiver-specific requirement since the action is flexible.
+                    bodyTypeMetadata = receiverBodyTypeMetadata;
+                }
+                else if ((~receiverBodyTypeMetadata.BodyType & bodyTypeMetadata.BodyType) != 0)
+                {
+                    // Failed subset check that WebHookMetadataProvider could not perform: Attribute must require the
+                    // same body type as receiver's metadata service or a subset. That is, `bodyTypeMetadata.BodyType`
+                    // flags must not include any beyond those set in `receiverBodyTypeMetadata.BodyType`.
+                    _logger.LogCritical(
```
Eilon: Also doesn't seem *critical*.

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
```diff
@@ -97,118 +128,153 @@ public void OnResourceExecuting(ResourceExecutingContext context)
...
+                        2,
+                        "Invalid '{MetadataType}.{PropertyName}' value '{PropertyValue}' in {AttributeType}. " +
+                        "This value must be equal to or a subset of the " +
+                        "'{ServiceMetadataType}.{ServicePropertyName}' value '{ServicePropertyValue}' for the " +
+                        "'{ReceiverName}' WebHook receiver.",
+                        typeof(IWebHookBodyTypeMetadata),
```
Eilon: These things that are `consts` should not be parameters in the log message because they are not data. They should just be in the string itself.

Eilon: Please review other log messages in this change and update as needed.

src/Microsoft.AspNetCore.WebHooks.Receivers/Metadata/IWebHookBodyTypeMetadataService.cs
```diff
@@ -1,14 +1,29 @@
...
     public interface IWebHookBodyTypeMetadataService : IWebHookBodyTypeMetadata, IWebHookReceiver
     {
+        /// <summary>
+        /// Gets the <see cref="WebHookBodyType"/> this receiver requires.
+        /// </summary>
+        /// <value>
+        /// <c>0</c> is not valid. Otherwise, any combination of defined flags is supported. Most receivers set a
+        /// single flag because they require a single body type.
+        /// </value>
+        new WebHookBodyType BodyType { get; }
```
Eilon: Why does this need a `new` property?

src/Microsoft.AspNetCore.WebHooks.Receivers/Properties/Resources.resx
```diff
@@ -141,27 +147,21 @@
   <data name="MetadataProvider_DuplicateMetadata" xml:space="preserve">
     <value>Duplicate '{0}' registrations found.</value>
   </data>
-  <data name="MetadataProvider_MissingAttributeMetadata" xml:space="preserve">
-    <value>'{0}' has a null {1} property but does not implement '{2}'.</value>
+  <data name="MetadataProvider_InvalidBodyType" xml:space="preserve">
+    <value>Invalid WebHook attribute '{0}' found.  Attributes implementing '{1}' must have valid {2} values. Must have a single {3} flag set or be '{4}'.</value>
```
Eilon: One space after a period.

src/Microsoft.AspNetCore.WebHooks.Receivers/Metadata/WebHookBodyType.cs
dougbu: Yup. Fortunately, this is the entire set of body types known in https://github.com/aspnet/AspNetWebHooks and I'm not aware of other senders that do anything different.

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
dougbu: Again, about to `throw` an `Exception` indicating the configuration is messed up. In this case, the `[GeneralWebHook]` attribute specified a `BodyType` this receiver doesn't support but the user didn't use `id`s to make the action unreachable.

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
dougbu: About to `throw` an `Exception` indicating the configuration is messed up. Nothing is wrong with the request but the application has a routable action that can't handle any any requests for this receiver.

Eilon: Yeah that isn't critical, it's just an error.

src/Microsoft.AspNetCore.WebHooks.Receivers/Metadata/IWebHookBodyTypeMetadataService.cs
Eilon: I don't really know the design, but doing a `new` property just to change the docs does not strike me as sound design. It's literally defining a new property and now there are two properties with the same name on types implementing this interface. That is clearly not the intention here.

dougbu: > clearly not the intention here

Agreed. And, looking again at the inherited `IWebHookBodyTypeMetadata.BodyType` doc comments, a lot doesn't apply here.

I'll separate the two interfaces.

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
dougbu: :ok:

src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs
dougbu: My intent was to avoid manual changes if the type name changes. Are you recommending I change e.g.
``` c#
"'{ServiceMetadataType}.{ServicePropertyName}' value '{ServicePropertyValue}' for the "
```
to
``` c#
$"'{typeof(IWebHookBodyTypeMetadataService)}.{nameof(IWebHookBodyTypeMetadataService.BodyType)}' value '{{ServicePropertyValue}}' for the "
```

dougbu:
:up::date: (rebased too)

dougbu:
@rynowak bump. I've addressed @Eilon's comments and would appreciate your input.

src/Microsoft.AspNetCore.WebHooks.Receivers/ApplicationModels/WebHookMetadataProvider.cs
```diff
@@ -144,168 +153,213 @@ private void Apply(ActionModel action)
...
+                default:
+                    {
+                        // Value contains undefined flags.
+                        var message = string.Format(
+                            CultureInfo.CurrentCulture,
+                            Resources.General_InvalidEnumValue,
+                            typeof(WebHookBodyType),
+                            actionBodyTypeMetadata.BodyType);
+                        throw new InvalidOperationException(message);
+                    }
```
rynowak: Do we really need all of this? This seem like it's really doing a ton of work to tell folks they did something silly.

src/Microsoft.AspNetCore.WebHooks.Receivers/Metadata/IWebHookBodyTypeMetadataService.cs
```diff
@@ -1,14 +1,27 @@
...
-    public interface IWebHookBodyTypeMetadataService : IWebHookBodyTypeMetadata, IWebHookReceiver
+    /// <remarks>
+    /// <see cref="ApplicationModels.WebHookModelBindingProvider"/> sets <see cref="Mvc.ModelBinding.BindingInfo"/>
+    /// properties based on this metadata or <see cref="IWebHookBodyTypeMetadata"/> and
+    /// <see cref="IWebHookBindingMetadata"/>. <see cref="Filters.WebHookVerifyBodyTypeFilter"/> confirms the request
+    /// body type based on this metadata. <see cref="Filters.WebHookEventMapperFilter"/> uses this metadata to decide
+    /// how to parse the request body and how to interpret
+    /// <see cref="IWebHookEventFromBodyMetadata.BodyPropertyPath"/>.
+    /// </remarks>
+    public interface IWebHookBodyTypeMetadataService : IWebHookMetadata, IWebHookReceiver
```
rynowak: I think I'm really confused about something and we should just discuss, what's the difference between this and `IWebHookBodyTypeMetadata`?

src/Microsoft.AspNetCore.WebHooks.Receivers/ApplicationModels/WebHookMetadataProvider.cs
rynowak: Commenting again after reading through the rest of the changes. About 60% of the code seems dedicated to detecting and reporting invalid cases. 

Is there a way we could simplify the contract that has to be implemented by the runtime - IE check once when actions are discovered, and then ignore/crash on invalid scenarios that can't be detected at startup?

dougbu: Discussed offline and I'll switch to support `[GeneralWebHook]` without resorting to a `[Flags]` enum.

dougbu:
:up::date: and ready for you @rynowak

dougbu:
730adf3128

PR #255, opened 2018-03-05
Clean up logging

dougbu:
- part of [discussion](https://github.com/aspnet/WebHooks/pull/254#discussion_r171360149) in #254
- don't log immediately before `throw`ing an `Exception`
- reduce log levels
- inline type names and other constants
- start event ids at `0` where possible


dougbu:
7598fe6721

Issue #260, opened 2018-03-12
Last refactor broke the Github example due to GeneralWebHook


WestDiscGolf:
After getting latest changes and re-running the Github sample, on start up an error occurs in `WebHookModelBindingProvider` in the `OnProviderExecuting` method.

It errors when registering the `FallbackHandler` from the github controller ...
```
[GeneralWebHook]
public IActionResult FallbackHandler(string receiverName, string id, string eventName)
```

The error occurs when it hits the following code ...
```
// Reachable only for [GeneralWebHook] cases. That attribute implements
// IWebHookBodyTypeMetadata and WebHookMetadataProvider passed it along.
bodyTypeMetadata = action.Properties[typeof(IWebHookBodyTypeMetadata)];
var actionBodyTypeMetadata = (IWebHookBodyTypeMetadata)bodyTypeMetadata;
bodyType = actionBodyTypeMetadata.BodyType;
```
as there is no property for the `typeof(IWebHookBodyTypeMetadata)` for the GeneralWebHook.

![image](https://user-images.githubusercontent.com/374343/37309337-89569eb6-2638-11e8-8589-3ee53e46163a.png)

It feels like there needs to be a GeneralMetadata class, which derives from WebHookMetadata, is registered in `WebHookServiceCollectionSetup`.

Thoughts?

WestDiscGolf:
Thinking maybe a general metadata class for the attribute to work from, or maybe a way to override the type it associated with.

Maybe an api like ...
```
[GeneralWebHook(Type = typeof(GitHubMetadata))]
```
... where you can override which type of meta data to use for the general web hook.

I've tried a hack of creating a GeneralMetadata class and registering it but didn't work and I'm missing something.

dougbu:
Thank you for your feedback. We're closing this issue as it is fixed in our working code base. More recent builds are available from https://dotnet.myget.org/gallery/aspnetcore-dev

This was a known temporary situation in 1e89a0fb21:
> temporarily **breaks GitHub** scenarios; real `IWebHookBodyTypeMetadataService` requirement coming in #254

Fixed in 730adf3128 (merged two days later).

WestDiscGolf:
I am experiencing the issue running the latest in the dev branch which has the referenced commit (https://github.com/aspnet/WebHooks/commit/730adf3128b60425157b53e248e1cea19dd74ad0) in it?

I've reset my local repository copy to make sure I have the latest ...
[...]

Issue #240, opened 2018-02-01
Add `ModelStateInvalidFilter` to actions with an applied `WebHookAttribute`


dougbu:
We should get rid of the `if (!ModelState.IsValid)`&hellip; boiler-plate in user code.

Should also `throw` an `InvalidOperationException` if `[ApiController]` is applied to a controller containing WebHook actions.

/cc @rynowak anything else we need to worry about now that WebHooks gets all the 2.1 goodness?

dougbu:
More code we can clean up using 2.1 features: Don't need to special-case media type suffix matches and should instead use `application/*+json` or `application/*+xml` in `WebHookVerifyBodyTypeFilter`. The following hack can go away: https://github.com/aspnet/WebHooks/blob/ca7f0e7c9a11663be336d0d0412e0e26b4d9117a/src/Microsoft.AspNetCore.WebHooks.Receivers/Filters/WebHookVerifyBodyTypeFilter.cs#L167-L171

mkArtakMSFT:
@danroth27, any objections about moving this out from preview2?

dougbu:
ec834ebabb

dougbu:
Merged to `dev` in 607d59b9e6

PR #267, opened 2018-03-19
Support receiver name / id lookup from `ActionDescriptor` in constraints

dougbu:
- pick up #240 and #260 fixes

### DIFF
--- a/src/Microsoft.AspNetCore.WebHooks.Receivers/Routing/WebHookRouteDataActionDescriptorExtensions.cs
+++ b/src/Microsoft.AspNetCore.WebHooks.Receivers/Routing/WebHookRouteDataActionDescriptorExtensions.cs
@@ -1,0 +1,95 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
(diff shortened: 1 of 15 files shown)


src/Microsoft.AspNetCore.WebHooks.Receivers/Routing/WebHookRouteDataActionDescriptorExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+
+using System;
+using Microsoft.AspNetCore.Mvc.Abstractions;
+using Microsoft.AspNetCore.Routing;
+
+namespace Microsoft.AspNetCore.WebHooks.Routing
+{
+    /// <summary>
+    /// Extension methods with <see cref="ActionDescriptor"/> parameters for the <see cref="RouteData"/> class.
+    /// </summary>
```
dougbu: Doc comments here mainly because I started with `WebHookRouteDataExtensions` and that's `public`. Was somewhat helpful when updating the constraints and may be helpful if we add or change constraints. But LMK if you prefer I remove these.

dougbu:
e4163deb1d in `release/2.1`
7545b61534 to merge this into `dev`

Issue #250, opened 2018-02-20
Support receiver name / id lookup in `ActionDescriptor` passed to constraints


dougbu:
To support frameworks built on top of ASP.NET Core WebHooks, make WebHook constraints aware of `ActionDescriptor.RouteValues`. This will enable higher-level frameworks to use route templates lacking `{webHookReceiver}` and `{id?}` parameter route values.

Suggest preferring `RouteData.Values` i.e. use `WebHookRouteDataExtensions.TryGetWebHookReceiverName(...)` and `TryGetWebHookReceiverId(...)` first. Fall back to `ActionDescriptor.RouteValues`.

(This extension is not necessary in WebHook filters because `ActionDescriptor.RouteValues` are added to `RouteData.Values` after the route is chosen.)

dougbu:
390ebbb258 in `release/2.1`
1aeaf69f87 merges this to `dev`

dougbu:
Ignore last comment

dougbu:
e4163deb1d in `release/2.1`
7545b61534 to merge this into `dev`

Issue #180, opened 2017-11-11
Add tests of ASP.NET Core receivers


dougbu:
Follow-on work item for #5.

dougbu:
@mkArtakMSFT cost is arbitrary. Will create a new task for work that doesn't fit in this milestone.

Aim for this milestone is to _at least_ have examples of tests in each area in the product i.e. application model providers, constraints, filters, and maybe attributes. Also need to show the pattern for functional tests of the samples.

mkArtakMSFT:
Reducing the cost to fit into this milestone. The rest of the work will happen in the future milestones.

dougbu:
Made some progress on this in Preview 2. Moving rest (including functional tests!) to RC 1.

dougbu:
Increased cost in hopes we can make more headway in this milestone.

WestDiscGolf:
@dougbu Are all the items in the 1.0.0-rc1 milestone the items which will make the release for when aspnetcore 2.1 is RTM'd? Are there others?

I am assuming this library is on the same release cycle ...

dougbu:
@WestDiscGolf of the still-open 1.0.0 RC1 issues, #275 should be included in the RC1 release. #180 is about engineering debt and, though I'm working on it now, isn't destined for our 2.1 branches.

Bottom line, the milestone of this issue isn't correct. We'll decide on the correct way to handle it later today.

WestDiscGolf:
@dougbu thanks for letting me know. Looking forward to the nuget release so I can release the one I've been working on.

Are any of the tests up for grabs? If I can find some time I'd like to contribute :-)

dougbu:
> Are any of the tests up for grabs?

Not at the moment. I may reach out to you (here) if I can't find sample messages for all services. But, coordination at the sub-issue level will get complicated. And, I'm making good progress 😄

WestDiscGolf:
OK, no worries :-)

dougbu:
Part 1 of test-only commits in 0eaf1540b8

dougbu:
Part 2 of test-only commits in d9c0651482. 65dcad481b merged that to `dev`.

PR #268, opened 2018-03-22
Merge into `dev`: Support receiver name / id lookup from `ActionDescriptor` in constraints

dougbu:
- #250
- add `WebHookRouteDataActionDescriptorExtensions` supporting these lookups
- add a bunch o' constraint tests (part of #180)

nit: leave `out` parameters `null` when `WebHookRouteDataExtensions` methods return `false`


dougbu:
FYI only

dougbu:
7545b61534

Issue #258, opened 2018-03-07
Reduce spread of `[GeneralWebHook]` support


dougbu:
Most of the WebHook constraints and filters do runtime look-ups (Linq queries) to find their metadata. This is decentralized even though it's not required for anything but actions with `[GeneralWebHook]` applied. Thought is to instead apply specific constraints and filters when processing the application model.

For the `[GeneralWebHook]` case, use different constraint and filter constructors that pass a new `interface`. That interface would hide the runtime lookup behind its facade. When processing a request, constraints and filters would use the interface to grab the metadata they need, then proceed in the normal (receiver-specific) way.

Should _not_ repeat any requirements the application model code enforces in the code for the `[GeneralWebHook]` case. Instead assume the retrieved metadata is consistent and let the constraints and filters "blow up" if not. This avoids additional code duplication and complexity.

dougbu:
Self-assigning because this needs to be done in the Preview2 milestone. Will otherwise be locked into the current approach.

dougbu:
@mkArtakMSFT doing this may reduce the size of other tasks and issues in this repo a bit. Issue is itself on the S / M border because it should mostly be mechanical.

dougbu:
BTW this is tied to main ASP.NET Core WebHooks 1.0.0 rock because `[GeneralWebHook]` actions are central to that.

WestDiscGolf:
Any update on the above and associated pull request? is this the last large change which will be going in before the full aspnet core 2.1 release?

Cheers

dougbu:
> Any update on the above and associated pull request?

Have been doing other things but will get back to this soon.
> is this the last large change which will be going in before the full aspnet core 2.1 release?

Not necessarily. Why do you ask?

WestDiscGolf:
Thanks for the update. I've been working on a Webhook that I want to nuget up but have been developing it in this solution to aid with debugging and getting the latest changes as the library has been developing. Just curious how much I'd have to tweak before I can get a preview out based on a stable nuget :-)

dougbu:
Reduced cost to track the review process that remains in this milestone.

dougbu:
c9ab0b4ede

PR #270, opened 2018-03-30
Reduce spread of `[GeneralWebHook]` support, stop adding global filters

dougbu:
- #258
- centralize and minimize metadata lookups, especially per-request lookups
- add `abstract` `WebHookMetadataProvider` and implementation (`DefaultWebHookMetadataProvider`)
- add core constraints and filters to actions from `IApplicationModelProvider` implementations, not globally
  - constraints and filters use `WebHookMetadataProvider` in `[GeneralWebHook]` scenarios
- add receiver-specific filters to actions from `WebHookActionModelFilterProvider` or `WebHookFilterProvider`
  - pattern is somewhat simpler than that for core filters; these filters need to check their own applicability
  - add `WebHookFilterProvider`, an `IFilterProvider` implementation
  - add `IWebHookFilterMetadata` to ease registration of receiver-specific filters
- implement `IOrderedFilter` in all filters
- consolidate and rename lots of classes
  - give `IApplicationModelProvider` implementations clearer, more specific names

nits:
- make doc comments more consistent
- correct a few oddly-named fields and variables


src/Microsoft.AspNetCore.WebHooks.Receivers/ApplicationModels/WebHookActionModelFilterProvider.cs
```diff
@@ -107,25 +113,39 @@ private void Apply(ActionModel action)
...
+                filterMetadataObject is IWebHookFilterMetadata filterMetadata)
+            {
+                var context = new WebHookFilterMetadataContext(action);
+                filterMetadata.AddFilters(context);
+                foreach (var filter in context.Results)
+                {
+                    filters.Add(filter);
+                }
+            }
+        }
```
rynowak: Can this be simplified further? Who's adding to `action.Properties`? Can we just add the filter then?

src/Microsoft.AspNetCore.WebHooks.Receivers/ApplicationModels/WebHookActionModelPropertyProvider.cs
```diff
@@ -50,60 +55,62 @@ public class WebHookMetadataProvider : IApplicationModelProvider
         /// <param name="verifyCodeMetadata">
         /// The collection of <see cref="IWebHookVerifyCodeMetadata"/> services.
         /// </param>
-        public WebHookMetadataProvider(
+        public WebHookActionModelPropertyProvider(
```
rynowak: I think we should spend some time talking through this. This is more indirection then I expected to see here and I'd like to understand it.

src/Microsoft.AspNetCore.WebHooks.Receivers/ApplicationModels/WebHookActionModelFilterProvider.cs
dougbu: `WebHookActionModelPropertyProvider` has been adding to `ActionModel.Properties` forever. It does that mainly to provide inputs for the other 3 `IApplicationModelProvider` implementations, including this one. (I added a read of `ApplicationDescriptor.Properties` in `WebHookFilterProvider` but that's unique and should remain so.)

I'd rather not remove the use of `ActionModel.Properties` because it avoids duplicated code across these providers. But, as you suggest below, let's chat about it.

dougbu: In an offline conversation, we decided this approach was unusual but fine.

src/Microsoft.AspNetCore.WebHooks.Receivers/ApplicationModels/WebHookActionModelPropertyProvider.cs
dougbu: As above, decided to leave this as-is.

dougbu:
:up::date:
- rebased on `dev`
- no longer in design

src/Microsoft.AspNetCore.WebHooks.Receivers/Metadata/WebHookMetadataProvider.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// <param name="receiverName">The name of an available <see cref="IWebHookReceiver"/>.</param>
+        /// <returns>
+        /// If the metadata exists, the <see cref="IWebHookBodyTypeMetadataService"/> for the receiver with given
+        /// <paramref name="receiverName"/>; <see langword="null"/> otherwise.
+        /// </returns>
+        /// <remarks>
+        /// Does not throw on missing metadata though an <see cref="IWebHookReceiver"/> lacking a registered
+        /// <see cref="IWebHookBodyTypeMetadataService"/> implementation is invalid. That error is handled elsewhere.
+        /// </remarks>
+        public abstract IWebHookBodyTypeMetadataService GetBodyTypeMetadata(string receiverName);
```
rynowak: Why is this one named `FooService` and the rest are not?

dougbu:
> I still think that what's going on here with `ActionDescriptor.Properties` is much much more trouble than it's worth, but I also don't think it's a priority to change

Agree this approach has downsides and that reworking these details shouldn't be a priority. Deciding on then changing to another approach (e.g. merging the `IApplicationModelProvider`s into one and passing context around) would take time better spent elsewhere.

dougbu:
c9ab0b4ede

PR #271, opened 2018-04-02
React to Mvc changes

pranavkm:


test/Microsoft.AspNetCore.WebHooks.Receivers.Test/ApplicationModels/WebHookMetadataProviderTests.cs
```diff
@@ -497,9 +494,8 @@ public void OnProvidersExecuting_SucceedsWithValidAttributesAndMetadata()
         {
             // Arrange
             var provider = new TestMetadataProvider(ValidMetadata);
-            var context = new ApplicationModelProviderContext(new[] { typeof(SomeController).GetTypeInfo() });
-            var defaultProvider = new DefaultApplicationModelProvider(Options.Create(new MvcOptions()));
-            defaultProvider.OnProvidersExecuting(context);
+            var controllerType = typeof(SomeController);
```
dougbu: What is this variable for? Inline the `typeof(...)` as you've done elsewhere.

test/Microsoft.AspNetCore.WebHooks.Receivers.Test/ApplicationModels/WebHookMetadataProviderTests.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.ApplicationModels;
 using Microsoft.AspNetCore.Mvc.Internal; // For DefaultApplicationModelProvider
```
dougbu: Side question: Do we have an issue about making it easy (or easier) to test `IApplicationModelProvider`s _without_ using `DefaultApplicationModelProvider`? That does most of the required setup, unfortunately.

Alternative would be an issue to make `DefaultApplicationModelProvider` public.

pranavkm: We don't, although that sounds like a great idea. I have similar code in NSwag and it felt icky to use an internal type in a non-Mvc project.

Issue #248, opened 2018-02-12
ApiController + binding on WebHook actions = sadness


pranavkm:
Consider:

```C#
[ApiController]
public class WebHookController : ControllerBase
{
    [StripeWebHook]
    public IActionResult Post(string id, string notificationId, string @event)
    {
        
    }
}
```

None of the 3-values get bound. This is because `ApiController` assumes these parameters come from query (they don't exist in route) and consequently assigns a `FromQuery` attribute on it. We should either
a) Make the behavior on ApiController less stringent (not set a binding source constraint on parameters)
b) Make WebHook set up the right metadata for this to work.

dougbu:
Let's be clear here: Were you sad or frustrated? I'd be frustrated!

dougbu:
BTW this issue relates to #240. WebHooks needs a few adjustments both to use some new 2.1 features e.g. `ModelStateInvalidFilter` and to avoid conflicts with other new 2.1 features e.g. `[ApiController]`. I like splitting the two sides of the coin: Handle the conflicts in this issue and think about using more new features in #240.

dougbu:
Inclinations:
1. `throw` an `InvalidOperationException` if mix of `[ApiController]` and a WebHook action is detected.
2. Is it possible to ensure the WebHook application model changes "win" for the WebHook action. May need a new "I do not compose" flag.
3. Enhance WebHook conventions to do the **Right Thing:tm:** when `[ApiController]` is applied. Do not require `[ApiController]` but allow it.

Chose (1).

dougbu:
I am unable to reproduce this issue. @pranavkm's test case may be different from mine.

Or, perhaps something changed since Preview1 in this area. I tested in the `dev` branch but will retest using Preview1 bits.

## Expected behavior (what I see)
MVC's `ApiBehaviorApplicationModelProvider` disallows use of `[ApiController]` and other `IApiBehaviorMetadata` attributes with actions that are not attribute routed. That provider executes before `WebHookRoutingProvider` and `WebHookRoutingProvider` makes the action _appear_ to be attribute routed (and `throw`s if it already is attribute routed). That is, `ApiBehaviorApplicationModelProvider` should `throw` when it sees a controller with `[ApiController]` containing an action with any `[WebHook]` attribute.

### Details
The `InvalidOperationException` thrown during `IApplicationModelProvider` execution at startup is:
```
[...]

PR #279, opened 2018-04-13
`throw` if a WebHook action has constraints or is attribute-routed

dougbu:
- #248
- do not detect `[ApiController]` itself
  - `ApiBehaviorApplicationModelProvider` throws if that attribute used on non-attribute-routed actions

nits:
- update `WebHookSelectorModelProvider` methods to match constraints added
- correct name of resource used in `WebHookSelectorModelProvider` to match class


dougbu:
13da6ec8ac

PR #286, opened 2018-04-21
Add functional tests requiring a non-empty request body

dougbu:
- #180 part 2 of n
- focus on unsigned requests for receivers where we already had example messages
  - copied most messages from aspnet/AspNetWebHooks tests
  - copied `ResourceFile` from aspnet/Mvc
- extend `WebHookTestFixture` to support using a `TestLoggerFactory` in some tests
- fix `BitbucketPush` to allow `[MinLength(1)]` on `Changes` property when fun in .NET 4.6.1

Coming soon:
- tool to generate request signatures for Dropbox, GitHub, Pusher, Stripe, and Trello
  - test privately with real secret keys
- functional tests involving signature verification, including unused Stripe and Trello resources
- functional tests for other samples: DynamicsCRM, MailChimp and Slack


dougbu:
d9c0651482

PR #287, opened 2018-04-21
Merge to `dev`: Add functional tests requiring a non-empty request body

dougbu:
- #180 part 2 of n
- focus on unsigned requests for receivers where we already had example messages
  - copied most messages from aspnet/AspNetWebHooks tests
  - copied `ResourceFile` from aspnet/Mvc
- extend `WebHookTestFixture` to support using a `TestLoggerFactory` in some tests
- fix `BitbucketPush` to allow `[MinLength(1)]` on `Changes` property when fun in .NET 4.6.1


dougbu:
FYI only. Will merge once CI checks succeed.

PR #298, opened 2018-05-04
Upgrade to netcoreapp22

ryanbrandenburg:


test/Directory.Build.props
```diff
@@ -2,9 +2,9 @@
   <Import Project="..\Directory.Build.props" />
 
   <PropertyGroup>
-    <DeveloperBuildTestTfms Condition=" '$(DeveloperBuildTestTfms)' == '' ">netcoreapp2.1</DeveloperBuildTestTfms>
+    <DeveloperBuildTestTfms Condition=" '$(DeveloperBuildTestTfms)' == '' ">netcoreapp2.2</DeveloperBuildTestTfms>
     <StandardTestTfms>$(DeveloperBuildTestTfms)</StandardTestTfms>
-    <StandardTestTfms Condition=" '$(DeveloperBuild)' != 'true' ">netcoreapp2.1</StandardTestTfms>
```
dougbu: Why delete this line? The [equivalent line](https://github.com/aspnet/WebHooks/pull/298/files#diff-6d410058c0601ff0430bd2efc448fa5eR9) in samples/Directory.Build.props (for `$(StandardTestWebsiteTfms)`) remains.

ryanbrandenburg: Feel free to clean more things if you want, but when I have to make changes to 50 repos like this I only make changes which 1) Apply to every repo or 2) Are required to keep things working/correct. If I do otherwise it turns into a huge time sink.

dougbu: I wasn't suggesting an additional change. Instead, why make this change at all (when it's not required for samples)?

ryanbrandenburg: I changed it at the time because @pranavkm wanted it removed because its a noop now. At this point we've already merged a ton of these in and leaving it as is would just be another way that our repos are different from each other.

Issue #318, opened 2018-07-25
Need to either disable Global Routing or "onboard" this repo


dougbu:
Tests are failing due to `AmbiguousMatchException`s which likely relate to Global Routing. Tests are about to be disabled but we should fix the product code.

See #313 #314 #315 #316 #317

dougbu:
Revert fad82fdb23 when this bug is fixed

Eilon:
@dougbu do we know that it's a bug in WebHooks product code or could it be an issue in Global Routing? (cc @rynowak @JamesNK @kichalla )

rynowak:
I've thought about this more, and I'm treating this as a bug in routing. Given our other discussions so far related to 2.2's routing features, I plan to make changes to MVC that will make webhooks 'just work'

dougbu:
@rynowak should we close this issue (maybe label it `external`) and open something similar in aspnet/Routing? The new bug could link to aspnet/WebHooks@bedc118aef as a commit to revert when it's fixed.

rynowak:
Sure, feel free to link to this one. https://github.com/aspnet/Routing/issues/630 I've added verifying that webhooks works with global routing to that issue's checklist.

dougbu:
Closing as described above. aspnet/Routing#630 covers moving aspnet/WebHooks into the new world.

Issue #317, opened 2018-07-25
Flaky test: WebHookAction_NoUUID_IsBadRequest


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=514940&buildTypeId=XPlat_Mac_MacUniverseCoreCLR) occasionally with the following error:
```
Microsoft.AspNetCore.Routing.Matchers.AmbiguousMatchException : The request matched multiple endpoints. Matches: 
BitbucketStronglyTypedCoreReceiver.Controllers.BitbucketController.BitbucketForIt (BitbucketStronglyTypedCoreReceiver)
BitbucketStronglyTypedCoreReceiver.Controllers.BitbucketController.BitbucketForPush (BitbucketStronglyTypedCoreReceiver)
BitbucketStronglyTypedCoreReceiver.Controllers.BitbucketController.Bitbucket (BitbucketStronglyTypedCoreReceiver)
   at Microsoft.AspNetCore.Routing.EndpointConstraints.EndpointConstraintEndpointSelector.SelectAsync(HttpContext httpContext, IEndpointFeature feature, CandidateSet candidates)
   at Microsoft.AspNetCore.Routing.Matchers.DfaMatcher.MatchAsync(HttpContext httpContext, IEndpointFeature feature)
   at Microsoft.AspNetCore.Routing.GlobalRoutingMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.TestHost.HttpContextBuilder.<>c__DisplayClass10_0.<<SendAsync>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.AspNetCore.TestHost.ClientHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Mvc.Testing.Handlers.CookieContainerHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Mvc.Testing.Handlers.RedirectHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Net.Http.HttpClient.FinishSendAsyncBuffered(Task`1 sendTask, HttpRequestMessage request, CancellationTokenSource cts, Boolean disposeCts)
[...]

Issue #315, opened 2018-07-25
Flaky test: WebHookAction_NoEventHeader_IsNotFound


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=514940&buildTypeId=XPlat_Mac_MacUniverseCoreCLR) occasionally with the following error:
```
Microsoft.AspNetCore.Routing.Matchers.AmbiguousMatchException : The request matched multiple endpoints. Matches: 
GitHubCoreReceiver.Controllers.GitHubController.HandlerForItsPushes (GitHubCoreReceiver)
GitHubCoreReceiver.Controllers.GitHubController.HandlerForIt (GitHubCoreReceiver)
GitHubCoreReceiver.Controllers.GitHubController.HandlerForPush (GitHubCoreReceiver)
GitHubCoreReceiver.Controllers.GitHubController.GitHubHandler (GitHubCoreReceiver)
GitHubCoreReceiver.Controllers.GitHubController.FallbackHandler (GitHubCoreReceiver)
   at Microsoft.AspNetCore.Routing.EndpointConstraints.EndpointConstraintEndpointSelector.SelectAsync(HttpContext httpContext, IEndpointFeature feature, CandidateSet candidates)
   at Microsoft.AspNetCore.Routing.Matchers.DfaMatcher.MatchAsync(HttpContext httpContext, IEndpointFeature feature)
   at Microsoft.AspNetCore.Routing.GlobalRoutingMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.TestHost.HttpContextBuilder.<>c__DisplayClass10_0.<<SendAsync>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.AspNetCore.TestHost.ClientHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Mvc.Testing.Handlers.CookieContainerHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Mvc.Testing.Handlers.RedirectHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Net.Http.HttpClient.FinishSendAsyncBuffered(Task`1 sendTask, HttpRequestMessage request, CancellationTokenSource cts, Boolean disposeCts)
[...]

Issue #314, opened 2018-07-25
Flaky test: WebHookAction_NonPost_IsNotAllowed


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=514940&buildTypeId=XPlat_Mac_MacUniverseCoreCLR) occasionally with the following error:
```
Microsoft.AspNetCore.Routing.Matchers.AmbiguousMatchException : The request matched multiple endpoints. Matches: 
SlackCoreReceiver.Controllers.SlackController.SlackForCommand (SlackCoreReceiver)
SlackCoreReceiver.Controllers.SlackController.SlackForTrigger (SlackCoreReceiver)
SlackCoreReceiver.Controllers.SlackController.Slack (SlackCoreReceiver)
   at Microsoft.AspNetCore.Routing.EndpointConstraints.EndpointConstraintEndpointSelector.SelectAsync(HttpContext httpContext, IEndpointFeature feature, CandidateSet candidates)
   at Microsoft.AspNetCore.Routing.Matchers.DfaMatcher.MatchAsync(HttpContext httpContext, IEndpointFeature feature)
   at Microsoft.AspNetCore.Routing.GlobalRoutingMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.TestHost.HttpContextBuilder.<>c__DisplayClass10_0.<<SendAsync>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.AspNetCore.TestHost.ClientHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Mvc.Testing.Handlers.CookieContainerHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Mvc.Testing.Handlers.RedirectHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Net.Http.HttpClient.FinishSendAsyncBuffered(Task`1 sendTask, HttpRequestMessage request, CancellationTokenSource cts, Boolean disposeCts)
   at Microsoft.AspNetCore.WebHooks.FunctionalTest.SlackCoreReceiverTest.WebHookAction_NonPost_IsNotAllowed(HttpMethod method) in /_/test/Microsoft.AspNetCore.WebHooks.FunctionalTest/SlackCoreReceiverTest.cs:line 54
[...]

PR #319, opened 2018-07-25
Disable flaky tests

dougbu:
- while we work in #318
- see flaky test reports #313, #314, #315, #316, and #317

nits:
- VS decided to "upgrade" the solution a bit


dougbu:
FYI only. Will merge once CI checks complete.

